PR #1, opened 2021-06-22
🎨 HPC-7936: Introduce new model abstraction lib

s0:
This PR moves the proposed functionality for the abstraction library into a new shared library that can be used across both the old and new API backend.

Differences with the original PR (other than being organized in its own repo): https://github.com/UN-OCHA/hpc_service/pull/2241

* The versioned model library has also been implemented using this new abstraction library, and all models migrated
* The way IDs work has been changed somewhat to directly use `io-ts` Types.

Still TODO:

* Documentation / guidance on creating new models (as requested by @rgautam98)


s0:
@ashwinkjoseph and @psneo7, i haven't got you access to everything yet, including the ticket information (this is on my list for this week), but would you like to give this PR an initial cursory review?

s0:
This is now ready.

@rgautam98 can you take a look at the included documentation (in `CONTRIBUTING.md`) to check if you feel this is sufficient?

package.json
```diff
@@ -0,0 +1,38 @@
...
+    "check-types": "tsc --noEmit",
+    "prepare": "husky install",
+    "lint-prettier": "prettier -c .",
+    "lint-eslint": "eslint --quiet .",
+    "lint": "yarn lint-prettier && yarn lint-eslint"
+  },
+  "dependencies": {
+    "fp-ts": "^2.6.6",
+    "io-ts": "2.2.9",
+    "knex": "0.21.1",
```
Pl217, line 17: Some of these packages can be updated to their latest version, but that could be done in separate PR.

Knex went from `0.21.17` to `0.95.0` and I didn't bother to find out why, but it always seems drastically outdated to me for that reason :)

Pl217 (approved):
I have made some changes to the commit history, hope you don't mind:
- Squashed one commit that adds newline (because prettier complained :)
- Fixed a typo in `CONTRIBUTING.md`
- Cemented prettier version, to use exact version and not "compatible" version (with caret ^)

I will assign @astroash and @rgautam98 for their respective reviews.

package.json
s0: We'll want to check if `hpc_service` still works with this version too then before updating here. I have a feeling that we may need to update things in a few places if so.

s0:
Merging to unblock other work

PR #5, opened 2021-07-14
Add governing entity model

Pl217:
Having governing entity in hpc-api-core models is needed for governing entity to be added to deferred fetchers in https://github.com/UN-OCHA/hpc_service/pull/2219

### DIFF
--- a/src/db/models/governingEntity.ts
+++ b/src/db/models/governingEntity.ts
@@ -2,6 +2,9 @@ import * as t from 'io-ts';
 
 import { brandedType } from '../../util/io-ts';
 import { Brand } from '../../util/types';
+import { defineLegacyVersionedModel } from '../util/legacy-versioned-model';
+import { ENTITY_PROTOTYPE_ID } from './entityPrototype';
+import { PLAN_ID } from './plan';
 
 export type GoverningEntityId = Brand<
   number,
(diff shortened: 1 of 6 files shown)


src/db/models/governingEntity.ts
```diff
@@ -12,3 +15,25 @@ export type GoverningEntityId = Brand<
 export const GOVERNING_ENTITY_ID = brandedType<number, GoverningEntityId>(
   t.number
 );
+
+export default defineIDModel({
```
s0: When running the unit tests for `hpc_service`, it fails with the message:

```
Table governingEntity is missing columns in definition: currentVersion, latestVersion, latestTaggedVersion, versionTags
```

```suggestion
export default defineLegacyVersionedModel({
```

src/db/models/governingEntity.ts
```diff
@@ -12,3 +15,25 @@ export type GoverningEntityId = Brand<
...
+  fields: {
+    generated: {
+      id: {
+        kind: 'branded-integer',
+        brand: GOVERNING_ENTITY_ID,
+      },
+    },
+    optional: {},
+    required: {
+      planId: { kind: 'branded-integer', brand: PLAN_ID },
```
s0, line 29: This should be under `accidentallyOptional`, it's nullable (according to the SQL schema).

src/db/models/governingEntity.ts
```diff
@@ -12,3 +15,25 @@ export type GoverningEntityId = Brand<
...
+export default defineIDModel({
+  tableName: 'governingEntity',
+  fields: {
+    generated: {
+      id: {
+        kind: 'branded-integer',
+        brand: GOVERNING_ENTITY_ID,
+      },
+    },
+    optional: {},
```
s0:
```suggestion
```

s0 (changes requested):

s0 (approved):

PR #11, opened 2021-08-13
Define `createBrandedValue` method

Pl217:
~~Move (and adapt) authentication and logging related code to the core repository, where both new (https://github.com/UN-OCHA/hpc-api) and old (https://github.com/UN-OCHA/hpc_service) APIs can use it.~~

This PR is repurposed to do some basic codebase maintenance and define `createBrandedValue` method needed for https://github.com/UN-OCHA/hpc_service/pull/2356


s0:
@Pl217 Have you got an associated PR for `hpc_service` in the works? Rather than reviewing this in isolation, i'd like to test it with the old API to make sure that everything still works

Pl217:
> @Pl217 Have you got an associated PR for `hpc_service` in the works? Rather than reviewing this in isolation, i'd like to test it with the old API to make sure that everything still works

I'm still working on that one. Since it requires much more changes and diligence, I pushed this PR and hpc-api related PR, which were already done -  https://github.com/UN-OCHA/hpc-api/pull/21

src/auth/role.data.ts
```diff
@@ -0,0 +1,178 @@
+/**
+ * The raw data from the following tables from prod:
+ *
+ * * `role`
+ * * `rolePermittedAction`
+ *
+ * @deprecated
+ */
+export const ROLES = [
```
s0: hmm.

While I appreciate why you moved this module over, it' very much an artifact of the previous authentication system, and we probably don't want to move it over at all. (I probably named it in a confusing way, apologies about that). The way it handles the concept of roles is very different from `roles.ts` where the roles are expressed as code rather than data in the database. This file is effectively a hard-coded snapshot of the expected roles from the old auth system from the database.

src/db/index.ts
```diff
@@ -21,10 +26,15 @@ export default (conn: Knex) => ({
...
   authTarget: authTarget(conn),
+  authToken: authToken(conn),
   form: form(conn),
   governingEntity: governingEntity(conn),
   operation: operation(conn),
   operationCluster: operationCluster(conn),
+  participant: participant(conn),
+  participantRole: participantRole(conn),
+  role: role(conn),
+  rolePermittedAction: rolePermittedAction(conn),
```
s0: we don't want to port these models as they are from the legacy authentication system that we're phasing out

src/Request.ts
```diff
@@ -0,0 +1,44 @@
+import Participant from './db/models/participant';
+import { InstanceOfModel } from './db/util/types';
+import * as hapi from '@hapi/hapi';
+import * as restify from 'restify';
+
+interface ExpandedRequest {
```
s0: Hmm this class also feels like something that's a little too specific to the way that we expect requests to be constructed for restify using our middlewear in `hpc_service` only, and not something that feels generally applicable for the legacy and new API.

The general library should not need to be aware of specific implementations / libraries that we're using to implement the rest API servers (and therefore should not need to have any restify / hapi types included at all). Instead, any interfaces should only really include stuff that's common across both usages.

src/Context.ts
```diff
@@ -0,0 +1,19 @@
...
+    mode?: string;
+    color?: boolean;
+  };
+  name?: string;
+}
+
+export default interface Context {
+  connection: Knex<any, unknown[]>;
+  config: Config;
+  request: Request;
```
s0: i would not pass the raw request as part of the context. for the boundary of this library (i.e. the data that we'll be giving to this library in any functions that we call), we'd need the data to already be abstracted in a way that is general and useful for the library to use. the library shouldn't need to know that it's supporting a request for restify, hapi, whatever etc... all it should know is, this is how i log stuff, this is my database connection, etc...

src/auth/index.ts
```diff
@@ -0,0 +1,683 @@
...
+ * the database using the tables permittedAction and rolePermittedAction.
+ *
+ * TODO: These tables should be removed,
+ * and the mapping should be stored in code instead.
+ *
+ * In the meantime, for new code requiring new actions,
+ * they can be added to this file and avoid the need for a migration or
+ * modifying roles in any way.
+ */
+type ExtendedPermittedAction = PermittedActionIdString | perms.GlobalPermission;
```
s0: we don't want to include any of the legacy auth stuff in what we port over.

src/auth/index.ts
```diff
@@ -0,0 +1,683 @@
...
+ * participantRole is tied to a specific object or not.
+ *
+ * @deprecated
+ */
+const ADDITIONAL_ACTIONS: {
+  [id: string]: ExtendedPermittedAction[] | undefined;
+} = {
+  hpcadmin: Object.values(perms.AUTH_PERMISSIONS.global),
+  planlead: ['canCreateOrganizations', 'viewCategories'],
+};
```
s0: for the library and the new API, it should only use / know about the new roles and actions

src/auth/index.ts
```diff
@@ -0,0 +1,683 @@
...
+  const addLegacyPermissions = async () => {
+    const globallyPermitted = await getLegacyGloballyPermittedActions(context);
+    if (globallyPermitted.size === 0) {
+      return;
+    }
+    allowed.global = allowed.global || new Set();
+    for (const permission of globallyPermitted) {
+      allowed.global.add(permission);
+    }
+  };
```
s0: this we will want to keep in `hpc_service` only, to support the endpoints that haven't been updated to use the new auth system only.

src/db/util/index.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Take a collection of objects,
+ * and create a map of them, using aparticular property as the key
+ */
+export const organizeObjectsByUniqueProperty = <I, P extends keyof I>(
```
s0: this function can apply more generally than jsut `db` related actions, so should probably live in a module in `src/util` instead.

src/db/util/async.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * Create an async function that can be called multiple times with different
+ * arguments, and result in a single invocation of an async function that
+ * serves all the given calls at once.
+ *
+ * This function takes advantage of the NodeJS single-threaded concurrency model
+ * to group multiple requests together using `setImmediate`.
+ */
+export const createGroupableAsyncFunction = <
```
s0: this function can apply more generally than jsut `db` related actions, so should probably live in a module in `src/util/async` instead.

src/logging/log.ts
```diff
@@ -0,0 +1,228 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/**
+ * Use of `any` in this module is generally deliberate to help with generics
+ */
+import bunyan = require('bunyan');
```
s0: It may be smarter to just expose an interface that can be served by bunyan, rather than adding bunyan as a direct dependency. i.e. port over LogContext etc... but the actual implementation for the logging can still be specific to the environment (i.e. `hpc_service`, `hpc-api`) etc... Which will give a bit more flexibility in terms of dependencies on these libraries, where `hpc-api` will likely be a bit more flexible.

src/util/error.ts
```diff
@@ -0,0 +1,114 @@
...
+  message: string;
+};
+
+export class ForbiddenError extends ExtendableError {
+  private readonly _originalError?: Error;
+  private readonly namespace: string;
+  private readonly body: ErrorBody;
+  private readonly statusCode: StatusCode;
+
+  constructor(message: string, namespace?: string, originalError?: Error) {
```
s0: let's completely get rid of the concept of namespace while we're at it. It causes our error + endpoint code to be so messy, and should be completely done for us automatically by the way of including stacktraces in our error logging.

s0 (changes requested):
There seem to be quite a few things that you've moved over here that I would have preferred for us to keep out of this codebase completely, in particular:

* legacy auth related stuff
* restify/hapi-specific implementation details / request objects, where i think we can do a much better job of abstracting out the concrete interfaces that the library actually needs to use.

Let's have a call to discuss what I have in mind.

Pl217:
Let's still keep the useful bits of this PR. There are some maintenance commits which I would like to keep and also define `createBrandedValue` which is needed for https://github.com/UN-OCHA/hpc_service/pull/2356.

s0 (approved):

PR #16, opened 2021-09-27
Define `projectVersion` and `projectVersionPlan` models

Pl217:


src/db/models/workflowStatusOption.ts
```diff
@@ -0,0 +1,15 @@
...
+export type WorkflowStatusOptionId = Brand<
+  number,
+  { readonly s: unique symbol },
+  'workflowStatusOption.id'
+>;
+
+export const WORKFLOW_STATUS_OPTION_ID = brandedType<
+  number,
+  WorkflowStatusOptionId
+>(t.number);
```
krishnareddy-tadi, line 15: Why didn't we added remaining fields planId,type and value in this model ?

src/db/models/project.ts
```diff
@@ -0,0 +1,12 @@
...
+import { brandedType } from '../../util/io-ts';
+import type { Brand } from '../../util/types';
+
+export type ProjectId = Brand<
+  number,
+  { readonly s: unique symbol },
+  'project.id'
+>;
+
+export const PROJECT_ID = brandedType<number, ProjectId>(t.number);
```
krishnareddy-tadi, line 12: Why didn't we added rest of the fields in this model ?

src/db/models/workflowStatusOption.ts
Pl217: Because I'm taking minimalistic approach here :)

I only need `projectVersion` and `projectVersionPlan` defined, but need these ID properties defined in order to reference them. The rest will be defined when there's a need. It's a usual approach we take in this repo, look at `attachmentPrototype.ts`

src/db/models/project.ts
Pl217: Because I'm taking minimalistic approach here :)

I only need `projectVersion` and `projectVersionPlan` defined, but need these ID properties defined in order to reference them. The rest will be defined when there's a need. It's a usual approach we take in this repo, look at `attachmentPrototype.ts`

krishnareddy-tadi (approved):
LGTM

PR #17, opened 2021-09-29
🔥 Remove `package-lock.json`

Pl217:
Since 3b726bceb48f01f9c7408f197ad47b4acc5243d3, package-lock.json is ignored in git, but should have also been removed, because it can lead to inconsistencies when two package managers are used.

From `yarn` log:
> "package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json"


Pl217:
@czmj Let's do a release after this is merged, so you can unblock https://github.com/UN-OCHA/hpc_service/pull/2412

rgautam98 (approved):
As we have the yarn.lock, we can remove the package-lock.json

PR #20, opened 2021-10-19
Add expiredData model

rgautam98:


rgautam98:
PR for migrations is at https://github.com/UN-OCHA/hpc_service/pull/2451

src/db/index.ts
```diff
@@ -35,4 +36,5 @@ export default (conn: Knex) => ({
   projectVersionPlan: projectVersionPlan(conn),
   reportingWindow: reportingWindow(conn),
   reportingWindowAssignment: reportingWindowAssignment(conn),
+  expiredData: expiredData(conn),
```
Pl217: Let's not break this alphabetical order

src/db/models/expiredData.ts
```diff
@@ -0,0 +1,31 @@
+import * as t from 'io-ts';
+import { DATE } from '../util/datatypes';
+import { defineRawModel } from '../util/raw-model';
+
+const expiredDataType = {
```
Pl217:
```suggestion
const EXPIRED_DATA_TYPE = {
```

src/db/models/expiredData.ts
```diff
@@ -0,0 +1,31 @@
...
+  solr: null,
+};
+
+const expiredObjectType = {
```
Pl217:
```suggestion
const EXPIRED_OBJECT_TYPE = {
```

Pl217 (changes requested):
Very minor comments.

I will need @s0 to verify that we don't need `createdAt` and `updatedAt` in this table, because usage of `defineRawModel` over `defineSequelizeModel` means just that.

If we don't need those fields, then you will need to remove them from migration in https://github.com/UN-OCHA/hpc_service/pull/2451, but you can address inline comments in a mean time :)

s0:
> I will need @s0 to verify that we don't need `createdAt` and `updatedAt` in this table, because usage of `defineRawModel` over `defineSequelizeModel` means just that.
> 
> If we don't need those fields, then you will need to remove them from migration in https://github.com/UN-OCHA/hpc_service/pull/2451, but you can address inline comments in a mean time :)

Yep we don't need those fields, given the temporary nature of the data in this table, that's not important =)

rgautam98:
Once this is approved, I'll update and push the migrations.

src/db/index.ts
rgautam98: I rearranged it in the order, but out of curiosity, why are we maintaining the order?

Pl217: I didn't start the order, but for me, it takes a little less effort to mentally parse such file if it's in some order, and alphabetical makes most sense.

I would say it improves readability in same manner that formatting the code does. That is why if you use command (`Ctrl + Shift + P`) in VS Code `Organize Imports` you will also get alphabetical order. Or maybe it was the easiest way for VS Code developers to write the logic :)

Anyhow, not only this usage needs to be in order, but its import as well. I'm gonna go ahead and make the change, so I can merge it, since you need this for https://github.com/UN-OCHA/hpc_service/pull/2451

Pl217 (approved):

PR #30, opened 2021-11-02
Define models needed for HPC-7904

Pl217:
There are a few changes I would request in #29, but it got already merged, so I made them here. Didn't want to open a new PR.

Took the opportunity to update Typescript, since I also updated it in https://github.com/UN-OCHA/hpc-api/pull/21.

Version bump is also done, because we'll need to do a release after this is merged, so that separate PR for version bump is not needed.


Pl217:
@s0 I would also like you to take a look, since I've made some changes to models defined in #29.
Also, some changes are relevant to my review in https://github.com/UN-OCHA/hpc_service/pull/2467

src/db/models/location.ts
```diff
@@ -10,3 +11,39 @@ export type LocationId = Brand<
...
+      // fetched as a string by knex, since it is bigint
+      validOn: { kind: 'checked', type: t.string },
+      parentId: { kind: 'branded-integer', brand: LOCATION_ID },
+    },
+    accidentallyOptional: {
+      adminLevel: { kind: 'checked', type: t.number },
+      status: {
+        kind: 'enum',
+        values: LOCATION_STATUS,
+      },
```
s0: `enum` should only be used when the property uses an enum defined in sql, this particular field is a string, so instead we should use a checked string with `t.keyof` to validate upon read, and restrict the type when writing.

The reason for this is that at some point I'd like to extend the validation function / unit tests to check that enum types are defined correctly, and allow the same values.

src/db/models/planVersion.ts
```diff
@@ -10,3 +14,43 @@ export type PlanVersionId = Brand<
...
+      currentReportingPeriodId: {
+        kind: 'branded-integer',
+        brand: PLAN_REPORTING_PERIOD_ID,
+      },
+      lastPublishedReportingPeriodId: { kind: 'checked', type: t.number },
+      // Even though this column isn't defined using DB enum only two values are used
+      clusterSelectionType: {
+        kind: 'enum',
+        values: PLAN_VERSION_CLUSTER_SELECTION_TYPE,
+      },
```
s0: should use a checked t.keyof type instead

s0 (changes requested):
2 minor changes need to be done. I'll do it quickly myself though as i'd like to get this in and have just reviewed

s0 (approved):

src/db/models/planVersion.ts
Pl217: Should have removed this comment above too. Since it "justifies" why `'enum'` was used

src/db/models/location.ts
Pl217: Makes sense even now, without enum definition checks

PR #31, opened 2021-11-03
🐛 Correct properties for attachment model

s0:
Stricter types can be used for objectType and type, which was the case
for the old attachment model, so we should continue that here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unocha/hpc-api-core",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "Core libraries supporting HPC.Tools API Backend",
   "license": "Apache-2.0",
   "private": false,
--- a/src/db/models/attachment.ts
+++ b/src/db/models/attachment.ts
@@ -17,10 +17,6 @@ export type AttachmentId = Brand<
 
 export const ATTACHMENT_ID = brandedType<number, AttachmentId>(t.number);
 
-export const COST_ATTACHMENT_VALUE = t.type({
-  cost: t.number,
-});
-
 export const ATTACHMENT_OBJECT_TYPE = t.keyof({
   governingEntity: null,
   plan: null,
--- a/src/db/models/attachment.ts
+++ b/src/db/models/attachment.ts
@@ -3,7 +3,10 @@ import { brandedType } from '../../util/io-ts';
 
 import type { Brand } from '../../util/types';
 import { defineLegacyVersionedModel } from '../util/legacy-versioned-model';
-import { ATTACHMENT_PROTOTYPE_ID } from './attachmentPrototype';
+import {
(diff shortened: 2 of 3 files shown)


Pl217 (changes requested):
Since, we're enriching attachment definition, I would take the opportunity to remove unused variable `COST_ATTACHMENT_VALUE` that is duplicated from `attachmentVersion.ts`, where it is actually used.  Both were added in f59ddc16959086e5c47ff6fe7a4a24eeafa69900

Can I take this PR to sneak in a commit that removes [this code comment](https://github.com/UN-OCHA/hpc-api-core/blob/develop/src/db/models/planVersion.ts#L47) since it [doesn't belong](https://github.com/UN-OCHA/hpc-api-core/pull/30#discussion_r741829569) after the changes we did today in 948d97d44d2e4eb8659953e824579ea3d62becf0?

Pl217:
Also, what are your plans for [HPC-8205 branch](https://github.com/UN-OCHA/hpc-api-core/tree/HPC-8205) that removes `COST_ATTACHMENT_VALUE` (which I asked above, before realizing this :smile:) and also makes changes to `attachmentPrototype` and `attachmentVersion`?

s0:
> Since, we're enriching attachment definition, I would take the opportunity to remove unused variable `COST_ATTACHMENT_VALUE` that is duplicated from `attachmentVersion.ts`, where it is actually used. Both were added in [f59ddc1](https://github.com/UN-OCHA/hpc-api-core/commit/f59ddc16959086e5c47ff6fe7a4a24eeafa69900)
> 
> Can I take this PR to sneak in a commit that removes [this code comment](https://github.com/UN-OCHA/hpc-api-core/blob/develop/src/db/models/planVersion.ts#L47) since it [doesn't belong](https://github.com/UN-OCHA/hpc-api-core/pull/30#discussion_r741829569) after the changes we did today in [948d97d](https://github.com/UN-OCHA/hpc-api-core/commit/948d97d44d2e4eb8659953e824579ea3d62becf0)?

sounds good to me

> Also, what are your plans for [HPC-8205 branch](https://github.com/UN-OCHA/hpc-api-core/tree/HPC-8205) that removes `COST_ATTACHMENT_VALUE` (which I asked above, before realizing this 😄) and also makes changes to `attachmentPrototype` and `attachmentVersion`?

That's still a WIP that i need to evolve a bit more, but i'll just rebase that ontop of whatever other changes we have in `hpc-api-core`, and will only finalize it when the related code in `hpc_service` is final and ready too, which I hope to have by end of day.

s0:
@Pl217 done

Pl217 (approved):
I have also bumped version here, because I will publish a new version once this PR is merged, since changes from #25 are needed for https://github.com/UN-OCHA/hpc_service/pull/2445

PR #32, opened 2021-11-03
✨ HPC-8205: Remaining changes neccesary for HPC-8205

s0:
Make remaining changes neccesary to support https://github.com/UN-OCHA/hpc_service/pull/2439

### DIFF
--- a/src/db/models/attachmentPrototype.ts
+++ b/src/db/models/attachmentPrototype.ts
@@ -37,15 +37,35 @@ const FIELDS = t.array(
 export const ATTACHMENT_PROTOTYPE_VALUE = t.intersection([
   // Required Fields
   t.type({
-    hasMeasures: t.number,
+    /**
+     * Some prototypes incorrectly quote numbers, which we need to account for
+     *
+     * TODO: clean prototype data and make type stricter
+     */
+    hasMeasures: t.union([
+      t.literal(0),
+      t.literal(1),
+      t.literal('0'),
+      t.literal('1'),
+    ]),
     name: LOCALIZED_STRING,
     entities: t.array(t.string),
   }),
   // Optional Fields
   t.partial({
     measureFields: FIELDS,
-    min: t.number,
-    max: t.number,
+    /**
+     * Some prototypes incorrectly quote numbers, which we need to account for
+     *
+     * TODO: clean prototype data and make type stricter
+     */
+    min: t.union([t.number, t.string]),
+    /**
+     * Some prototypes incorrectly quote numbers, which we need to account for
+     *
+     * TODO: clean prototype data and make type stricter
+     */
+    max: t.union([t.number, t.string]),
     metrics: FIELDS,
   }),
 ]);
--- a/src/db/models/json/attachment.ts
+++ b/src/db/models/json/attachment.ts
@@ -1,65 +1,65 @@
 import * as t from 'io-ts';
+import {
+  CASELOAD_VALUE as CASELOAD_ATTACHMENT_VALUE,
+  INDICATOR_VALUE as INDICATOR_ATTACHMENT_VALUE,
+} from './indicatorsAndCaseloads';
+
+const COST_ATTACHMENT_VALUE = t.intersection([
+  t.exact(
+    t.type({
+      cost: t.number,
+    })
+  ),
+  t.exact(
+    t.partial({
+      /**
+       * When necessary, a cost breakdown can be provided with respect to a
+       * particular collection of objects.
+       *
+       * For example, when the object of this attachment is a governing entity,
+       * a breakdown needs to be provided for each of the global clusters,
+       * (even if the governing entity has 0 global clusters)
+       * and the total sum of the breakdown must match the overall cost when
+       * non-empty.
+       */
+      breakdown: t.array(
+        t.exact(
+          t.type({
+            objectId: t.number,
+            cost: t.number,
+          })
+        )
+      ),
+      // TODO: delete these properties once we've confirmed that they're not
+      // needed for any code that reads cost attachments
+      // (they seem to be produced by RPM frontend code related to other types)
+      /** @deprecated TODO: remove (this doesn't look like it's needed for cost attachments) */
+      name: t.string,
+      /** @deprecated TODO: remove (this doesn't look like it's needed for cost attachments) */
+      description: t.string,
+      /** @deprecated TODO: remove (this doesn't look like it's needed for cost attachments) */
+      metrics: t.exact(
+        t.partial({
+          values: t.exact(t.partial({})),
+          measureFields: t.array(t.UnknownRecord),
+        })
+      ),
+    })
+  ),
+]);
 
 export const ATTACHMENT_VERSION_VALUE = {
-  /**
-   * TODO: Replace this with type that represents expected values
-   */
-  caseLoad: t.unknown,
+  caseLoad: CASELOAD_ATTACHMENT_VALUE,
   /**
    * TODO: Replace this with type that represents expected values
    */
   contact: t.unknown,
-  cost: t.intersection([
-    t.exact(
-      t.type({
-        cost: t.number,
-      })
-    ),
-    t.exact(
-      t.partial({
-        /**
-         * When necessary, a cost breakdown can be provided with respect to a
-         * particular collection of objects.
-         *
-         * For example, when the object of this attachment is a governing entity,
-         * a breakdown needs to be provided for each of the global clusters,
-         * (even if the governing entity has 0 global clusters)
-         * and the total sum of the breakdown must match the overall cost when
-         * non-empty.
-         */
-        breakdown: t.array(
-          t.exact(
-            t.type({
-              objectId: t.number,
-              cost: t.number,
-            })
-          )
-        ),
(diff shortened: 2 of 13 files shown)


src/db/util/validation.ts
```diff
@@ -104,7 +104,25 @@ interface DataValidator<T> {
    * Check that the given data has the expected type,
    * and if not throw an error explaining why
    */
-  validateAndFilter(data: unknown): T;
+  validateAndFilter(data: unknown, genIdentifier: () => string): T;
+}
+
+export class DataValidationError extends Error {
```
Pl217: Wow, thanks a lot for introducing this. It will be so much easier to debug data validation problems.

I think this error definition should live in `src/db/util/errors.ts`

src/util/io-ts.ts
```diff
@@ -17,3 +19,77 @@ export const brandedType = <Base, B extends Brand<Base, any, any>>(
...
+/**
+ * Print out io-ts decoding errors in a more user-readable format for use in
+ * error messages / debugging
+ */
+export const ioTsErrorFormatter = (decoded: Left<t.Errors>): string[] =>
+  decoded.left.map((error) => {
+    const path = error.context
+      // Filter out useless / context paths
+      .filter((c, i) => {
+        const parent: t.ContextEntry | undefined = error.context[i - 1];
```
Pl217, line 32: I just love that you don't have to worry about those key out of bound exceptions that other languages have. You get a nice `undefined` :)

I know this method is copied from hpc_service and don't forget to remove it from there. Same goes for `indexedObjectType` (and all other code migrated here :)

src/db/models/attachmentPrototype.ts
```diff
@@ -37,15 +37,35 @@ const FIELDS = t.array(
...
+     * Some prototypes incorrectly quote numbers, which we need to account for
+     *
+     * TODO: clean prototype data and make type stricter
+     */
+    hasMeasures: t.union([
+      t.literal(0),
+      t.literal(1),
+      t.literal('0'),
+      t.literal('1'),
+    ]),
```
Pl217, line 50: This is actually not representing what data **can** be stored. Maybe it is true for the data that **is stored already**, I haven't checked (btw, do you know any efficient way to query JSON columns in Postgres to retrieve patterns), and maybe possibly true for the data we **want** stored (except for string numbers).

On the other hand, RPM UI allows for any value as far as I can see. I just tried putting number 2 and letter 'a' and both got stored nicely. With validation, we will prevent such cases, but it should be handled on the UI side as well.

We just need to make sure that there are no outliers for this field, then write admin command and clean prototype data.

But, then again, this should actually be boolean. Even if we allowed for any number and only treat value 1 as true and all other values as false is wrong and has bitten us already.

So, I think the right approach would be to tackle this in a separate ticket. We should write admin command that does the "migration" (person who does this should also inspect the data to see whether there are any strange pieces of data), and moves all falsy values to `false` (not `'false'`!) and truthy values to `true` (not `'true'`). Other strategy might also be applied, really depends on how much invalid values we have for `hasMeasures`.

Then, we only need to update this type definition here and everything should be fine. RPM already uses a library (I think `jsoneditor`) that will display a nice check box for us.
The only problem I see with this approach is internationalization.

src/db/models/attachmentPrototype.ts
```diff
@@ -37,15 +37,35 @@ const FIELDS = t.array(
...
   }),
   // Optional Fields
   t.partial({
     measureFields: FIELDS,
-    min: t.number,
-    max: t.number,
+    /**
+     * Some prototypes incorrectly quote numbers, which we need to account for
+     *
+     * TODO: clean prototype data and make type stricter
```
Pl217, line 60: Also, if possible, make sure it no longer happens. It will be validated by v4 models, but old models will still allow for this mixture (or is this a relict of a past?). Once admin command is in there, it could be ran periodically maybe, but problem should be truly solved.

It could also be achieved if old model that works with this data is replaced by v4 model.

src/db/models/json/attachment.ts
```diff
@@ -0,0 +1,80 @@
...
+const COST_ATTACHMENT_VALUE = t.intersection([
+  t.exact(
+    t.type({
+      cost: t.number,
+    })
+  ),
+  t.exact(
+    t.partial({
+      /**
+       * When neccesary, a cost breakdown can be provided with respect to a
```
Pl217:
```suggestion
       * When necessary, a cost breakdown can be provided with respect to a
```

src/db/models/json/attachment.ts
```diff
@@ -0,0 +1,80 @@
...
+        })
+      ),
+    })
+  ),
+]);
+
+export const ATTACHMENT_VERSION_VALUE = {
+  caseLoad: CASELOAD_ATTACHMENT_VALUE,
+  /**
+   * TODO
```
Pl217: Please write something in this TODO

`contact` seems to have following structure
```js
contact: t.type({
  leadAgency: t.string,
  contactName: t.string,
  contactEmail: t.string
})
```
but other keys are also available, like `'metrics'` and `'description'` :smile: like you defined for `COST_ATTACHMENT_VALUE`

src/db/models/json/attachment.ts
```diff
@@ -0,0 +1,80 @@
...
+  cost: COST_ATTACHMENT_VALUE,
+  /**
+   * TODO
+   */
+  fileWebContent: t.unknown,
+  indicator: INDICATOR_ATTACHMENT_VALUE,
+  /**
+   * TODO
+   */
+  textWebContent: t.unknown,
```
Pl217, line 66: Seems this needs to have `name` and `content?`, but of course we have `metrics` and `description` in DB, like for `contact`

src/db/models/json/indicatorsAndCaseloads.ts
```diff
@@ -0,0 +1,188 @@
+import * as t from 'io-ts';
```
Pl217, line 1: Some nice restructuring done in this file (compared to hpc_service) :clap:

Pl217 (changes requested):
Nice changes! Now that I look at my inline comments, maybe only a couple of them are actionable and rest are either just comments or talking about future actions we should take.

Let's not forget to refactor everything in hpc_service to what is migrated to hpc-api-core.

src/db/util/validation.ts
s0: Done, also took the opportunity to introduce `isDataValidationError` and set `this.name` in `ConcurrentModificationError`.

src/util/io-ts.ts
s0: Done in https://github.com/UN-OCHA/hpc_service/pull/2439 in `9cf3377c0e0551d7a6baffa42e26485d0d84244c`

src/db/models/json/attachment.ts
s0: I think i'd rather leave this as `t.unknown` until we have a need to actually read or write data to these types of attachments using the v4 model, at which point we can test it more thoroughly. FOr now I will still update the TODO to say `TODO: Replace this with type that represents expected values`

src/db/models/json/attachment.ts
s0: Same as above. I'll add a message to the TODO but leave it as `t.unknown`

src/db/models/attachmentPrototype.ts
s0: Agreed on all counts. Though i don't feel the internationalization issue is a problem, if a user is capable of writing a json prototype, and know json syntax, then they know about `true` and `false`.

And right, as you say any data can be stored right now, but at the very least this covers the majority of cases for current active data that's stored in RPM, which means that it can be used, and if any new invalid data is stored, then it will fail gracefully by reporting the invalid data and 500ing.

We definitely want the validation to occur upon saving / updating these attachments. IMO the validation can still be defined and occur server-side, as long as an error is propigated up to the frontend and displayed to the user somehow. So we want to update the 

Right now the v4 library doesn't do any runtime validation upon save (instead relying on the type safety of the caller), but if we wanted to start using this model from javascript endpoints (or any javascript data that doesn't go via a typescript `unknown` type forcing us to validate before sending it to the model), then we may also want to add runtime type validation to `create` and `update` in `raw-model.ts`.

Created an issue here to track: https://humanitarian.atlassian.net/browse/HPC-8258

s0:
Okay I think I've now addressed all actionable comments, back to you for review @Pl217

src/db/models/attachmentPrototype.ts
Pl217: Thank you for creating the ticket.

What do you think about doing data validation before inserting/updating in `create`/`update`? That would require validation of user data. What we do now is insert/update first and call `validateAndFilter` on instance created/updated.

While that is good for `find`, we maybe want to have runtime type validation for user data supplied. I am also suspicious of this, because we may need to update a lot of the code as it will reveal lots of bugs, since many json fields provide more data than needed. Also, there are many numbers stored as strings.

In `validateAndFilter`, `t.exact` does the filtering, but I think codec check would fail if we do this before actual save. What is your thinking on this?

> So we want to update the

This sentence was left incomplete :)

Pl217 (approved):

PR #33, opened 2021-11-03
♻️ Move HPC-8205 functions into hpc-api-core

s0:
Following on from #32

https://github.com/UN-OCHA/hpc_service/pull/2439 Introduces a number of general-purpose functions for reading and validating particular sets of data from the database that will likely need to be used when implementing certain functionality in the new v4 API. Rather than have them live in just hpc_service, given that they already only use the v4 models, they may as well live in this repo.

@Pl217 thoughts?

### DIFF
--- a/src/lib/data/governingEntities.ts
+++ b/src/lib/data/governingEntities.ts
@@ -102,7 +102,9 @@ const composeCustomReferenceForGoverningEntity = ({
 }): string => {
   const prototype = prototypes.get(governingEntity.entityPrototypeId);
   if (!prototype) {
-    throw new Error('Missing prototype');
+    throw new Error(
+      `Missing prototype for governingEntity ${governingEntity.id}`
+    );
   }
   return `${prototype.refCode}${governingEntityVersion.customReference}`;
 };
--- a/src/lib/data/governingEntities.ts
+++ b/src/lib/data/governingEntities.ts
@@ -1,0 +1,108 @@
+import { EntityPrototypeId } from '../../db/models/entityPrototype';
+import { GoverningEntityId } from '../../db/models/governingEntity';
+import { PlanId } from '../../db/models/plan';
+import { Database } from '../../db/type';
+import { InstanceDataOfModel } from '../../db/util/raw-model';
+import { organizeObjectsByUniqueProperty } from '../../util';
+
+/**
+ * A map from governing entity ids to an object that contains both the
+ * governingEntity and governingEntityVersion for each entry.
+ */
+export type MapOfGoverningEntities = Map<
+  GoverningEntityId,
+  {
+    governingEntity: InstanceDataOfModel<Database['governingEntity']>;
+    governingEntityVersion: InstanceDataOfModel<
+      Database['governingEntityVersion']
+    >;
+    customRef: string;
+  }
+>;
+
+export const getAllGoverningEntitiesForPlan = async ({
+  database,
+  planId,
+  prototypes,
+}: {
+  database: Database;
+  planId: PlanId;
+  version: 'latest';
+  /**
+   * A map of prototypes that **must** include the prototype for governing
+   * entities of this plan.
+   *
+   * Usually the full list of prototypes for a plan is passed in.
+   */
+  prototypes: Map<
+    EntityPrototypeId,
+    InstanceDataOfModel<Database['entityPrototype']>
+  >;
+}): Promise<MapOfGoverningEntities> => {
+  const ges = await database.governingEntity.find({
+    where: {
+      planId,
+    },
+  });
+  const gevs = await database.governingEntityVersion.find({
+    where: (builder) =>
+      builder
+        .whereIn(
+          'governingEntityId',
+          ges.map((ge) => ge.id)
+        )
+        .andWhere('latestVersion', true),
+  });
+  const gevsByGoverningEntityId = organizeObjectsByUniqueProperty(
+    gevs,
+    'governingEntityId'
+  );
+  const result: MapOfGoverningEntities = new Map();
+  for (const governingEntity of ges) {
+    const governingEntityVersion = gevsByGoverningEntityId.get(
+      governingEntity.id
+    );
+    if (!governingEntityVersion) {
+      throw new Error(
+        `Missing governing entity version for ${governingEntity.id}`
+      );
+    }
+    const customRef = composeCustomReferenceForGoverningEntity({
+      governingEntity,
+      governingEntityVersion,
+      prototypes,
+    });
+    result.set(governingEntity.id, {
+      governingEntity,
+      governingEntityVersion,
+      customRef,
+    });
+  }
+  return result;
+};
+
+const composeCustomReferenceForGoverningEntity = ({
+  governingEntity,
+  governingEntityVersion,
+  prototypes,
+}: {
+  governingEntity: InstanceDataOfModel<Database['governingEntity']>;
+  governingEntityVersion: InstanceDataOfModel<
+    Database['governingEntityVersion']
+  >;
+  /**
+   * A map of prototypes that **must** include the prototype for the entity.
+   *
+   * Usually the full list of prototypes for a plan is passed in.
+   */
+  prototypes: Map<
+    EntityPrototypeId,
+    InstanceDataOfModel<Database['entityPrototype']>
+  >;
+}): string => {
+  const prototype = prototypes.get(governingEntity.entityPrototypeId);
+  if (!prototype) {
+    throw new Error('Missing prototype');
+  }
+  return `${prototype.refCode}${governingEntityVersion.customReference}`;
+};
--- a/src/lib/data/planEntities.ts
+++ b/src/lib/data/planEntities.ts
@@ -92,12 +92,14 @@ export const getAndValidateAllPlanEntities = async ({
   for (const planEntity of planEntities) {
     const planEntityVersion = pevsByPlanEntityId.get(planEntity.id);
     if (!planEntityVersion) {
-      throw new Error('Missing plan entity version');
+      throw new Error(
+        `Missing plan entity version for planEntity ${planEntity.id}`
+      );
     }
 
     const prototype = prototypes.get(planEntity.entityPrototypeId);
     if (!prototype) {
-      throw new Error('Missing prototype');
+      throw new Error(`Missing prototype for planEntity ${planEntity.id}`);
     }
 
     const refAndType = getCustomReferenceAndTypeForPlanEntity({
@@ -216,20 +218,22 @@ const getCustomReferenceAndTypeForPlanEntity = ({
 } => {
   const prototype = prototypes.get(planEntity.entityPrototypeId);
   if (!prototype) {
-    throw new Error('Missing prototype');
+    throw new Error(`Missing prototype for planEntity ${planEntity.id}`);
   }
   const parent = entityAssociations.get(planEntity.id);
   let ref = '';
   if (parent) {
     const ge = governingEntities.get(parent.parentId);
     if (!ge) {
-      throw new Error('Missing governing entity');
+      throw new Error(
+        `Missing governing entity for planEntity ${planEntity.id}`
+      );
     }
     ref = `${ge.customRef}/`;
   }
   ref += `${prototype.refCode}${planEntityVersion.customReference}`;
   if (!prototype.refCode) {
-    throw new Error('Missing refCode');
+    throw new Error(`Missing refCode for entityPrototype ${prototype.id}`);
   }
   return {
     customRef: ref,
--- a/src/lib/data/planEntities.ts
+++ b/src/lib/data/planEntities.ts
@@ -61,10 +61,7 @@ export const getAndValidateAllPlanEntities = async ({
   const pevs = await database.planEntityVersion.find({
     where: (builder) =>
       builder
-        .whereIn(
-          'planEntityId',
-          planEntities.map((pe) => pe.id)
-        )
+        .whereIn('planEntityId', [...planEntityIDs])
         .andWhere('latestVersion', true),
   });
   const pevsByPlanEntityId = organizeObjectsByUniqueProperty(
@@ -75,10 +72,7 @@ export const getAndValidateAllPlanEntities = async ({
   const entitiesAssociation = await database.entitiesAssociation.find({
     where: (builder) =>
       builder
-        .whereIn(
-          'childId',
-          planEntities.map((pe) => pe.id)
-        )
+        .whereIn('childId', [...planEntityIDs])
         .andWhere('parentType', 'governingEntity')
         .andWhere('childType', 'planEntity'),
   });
--- a/src/lib/data/planEntities.ts
+++ b/src/lib/data/planEntities.ts
@@ -1,0 +1,238 @@
+import { EntityPrototypeId } from '../../db/models/entityPrototype';
+import { GoverningEntityId } from '../../db/models/governingEntity';
+import { PlanId } from '../../db/models/plan';
+import { PlanEntityId } from '../../db/models/planEntity';
+import { Database } from '../../db/type';
+import { InstanceDataOfModel } from '../../db/util/raw-model';
+import { organizeObjectsByUniqueProperty } from '../../util';
+import { MapOfGoverningEntities } from './governingEntities';
+
+export type ValidatedPlanEntity = {
+  id: PlanEntityId;
+  type: string;
+  customRef: string;
+  description: null | string;
+  /**
+   * List of any entities that this one "supports"
+   * (e.g. specific objectives support strategic objectives)
+   */
+  supports: PlanEntityId[];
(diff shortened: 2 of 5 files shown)


s0:
Still need to make the Errors more descriptive as requested in https://github.com/UN-OCHA/hpc_service/pull/2439#pullrequestreview-786825586

s0:
All errors now include IDs to help debug

src/lib/data/governingEntities.ts
```diff
@@ -0,0 +1,107 @@
...
+}): Promise<MapOfGoverningEntities> => {
+  const ges = await database.governingEntity.find({
+    where: {
+      planId,
+    },
+  });
+  const gevs = await database.governingEntityVersion.find({
+    where: (builder) =>
+      builder
+        .whereIn('governingEntityId', [...new Set(ges.map((ge) => ge.id))])
```
Pl217: Governing entity IDs already must be unique, due to DB primary key constraints, why put them through `Set`?
```suggestion
        .whereIn('governingEntityId', ges.map((ge) => ge.id))
```

src/lib/data/governingEntities.ts
```diff
@@ -0,0 +1,107 @@
...
+  prototypes: Map<
+    EntityPrototypeId,
+    InstanceDataOfModel<Database['entityPrototype']>
+  >;
+}): string => {
+  const prototype = prototypes.get(governingEntity.entityPrototypeId);
+  if (!prototype) {
+    throw new Error(
+      `Missing prototype for governingEntity ${governingEntity.id}`
+    );
```
Pl217, line 107: I was looking at a version before latest rebase (had it pulled locally), which had old message, without the ID information. Thanks for adding this additional error message information.

I was also thinking whether we can create a function to wrap around this pattern that we commonly use, but displaying useful error message will require that method to be really ugly, and not actually provide real benefit.

We've seen this so many times:
- Use `Map.prototype.get()`
- Check for possible `undefined` value 
- `throw new Error()` with mention of an ID

I was also thinking whether we should wrap around `Map` object and provide info like `name` when creating `Map` instance. We could then use `name` to say `Missing ${_name} for ID ${id}`, but that will also limit us in ways we want to display error messages. For example here, you search for entity prototype with `get(governingEntity.entityPrototypeId)`, but then you mention governing entity in error message, which would require our abstraction to be really messy to pull this out.

src/lib/data/planEntities.ts
```diff
@@ -0,0 +1,236 @@
...
+    EntityPrototypeId,
+    InstanceDataOfModel<Database['entityPrototype']>
+  >;
+}): Promise<ValidatedPlanEntities> => {
+  const planEntities = await database.planEntity.find({
+    where: {
+      planId,
+    },
+  });
+  const planEntityIDs = new Set(planEntities.map((pe) => pe.id));
```
Pl217, line 59: You don't need `Set` to make this a structure of unique plan entity IDs, since uniqueness is already enforced by DB primary key constraints.

However, `Set.prototype.has()` is nicer (and I think faster) way to search compared to `Array.prototype.indexOf()`, so we can keep it.

src/lib/data/planEntities.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  const pevs = await database.planEntityVersion.find({
+    where: (builder) =>
+      builder
+        .whereIn('planEntityId', [...new Set(planEntities.map((pe) => pe.id))])
```
Pl217: No need to map more than once, we already have this set
```suggestion
        .whereIn('planEntityId', [...planEntityIDs])
```

src/lib/data/planEntities.ts
```diff
@@ -0,0 +1,236 @@
...
+  });
+  const pevsByPlanEntityId = organizeObjectsByUniqueProperty(
+    pevs,
+    'planEntityId'
+  );
+
+  const entitiesAssociation = await database.entitiesAssociation.find({
+    where: (builder) =>
+      builder
+        .whereIn('childId', [...new Set(planEntities.map((pe) => pe.id))])
```
Pl217: No need to map more than once, we already have this set
```suggestion
        .whereIn('childId', [...planEntityIDs])
```

src/lib/data/planEntities.ts
```diff
@@ -0,0 +1,236 @@
...
+        .andWhere('parentType', 'governingEntity')
+        .andWhere('childType', 'planEntity'),
+  });
+  const eaByChildId = organizeObjectsByUniqueProperty(
```
Pl217, line 79: By how DB tables are designed, this property isn't guaranteed to be unique. Having `entitiesAssociation` table suggests we have many-to-many relationship between governing entities and plan entites.

I have queried `entitiesAssociation` table on my local backup of production data snapshot and there are no plan entities with multiple governing entity parents. Making it possible for a plan entity to have only one governing entity parent should have been enforced in DB schema, and having `parentGoverningEntityId` column in `planEntity` table would be my way to go about this.

On the other hand, in the early days, there was maybe little knowledge about how this relationship would evolve, so they kept it flexible, and I'm wondering should we do anything about it, or just assume only a single governing entity parent.

src/lib/data/planEntities.ts
```diff
@@ -0,0 +1,236 @@
...
+      governingEntity: eaByChildId.get(planEntity.id)?.parentId || null,
+    };
+
+    // Use entity details if possible
+    if (planEntityVersion.value) {
+      // Able to inspect entity details
+      entityDetails.description = planEntityVersion.value.description;
+      const supportingPlanEntityIds = planEntityVersion.value.support
+        .map((s) => s.planEntityIds)
+        .flat();
```
Pl217: `Array.prototype.flatMap()` "is identical to a `map()` followed by a `flat()` of depth 1, but slightly more efficient than calling those two methods separately".

And `Array.prototype.flat()` accepts `depth` parameter, which defaults to 1 when not supplied.
```suggestion
        .flatMap((s) => s.planEntityIds);
```

src/lib/data/projects.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+  });
+  const pvpsByProjectVersionId = organizeObjectsByUniqueProperty(
+    pvps,
+    'projectVersionId'
+  );
+  const projectVersions = await database.projectVersion.find({
+    where: (builder) =>
+      builder.whereIn('id', [
+        ...new Set(pvps.map((pvp) => pvp.projectVersionId)),
```
Pl217, line 33: If `projectVersionId` is a unique property of `pvps`, why put this map through `Set`? :)

You could use keys of `pvpsByProjectVersionId`
```suggestion
        ...pvpsByProjectVersionId.keys(),
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
+import { isRight } from 'fp-ts/lib/Either';
```
Pl217: I've noticed all other files use plural names in this directory, except this one. What a nitpick :smile:

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+    type: t.literal('indicator'),
+    value: ATTACHMENT_VERSION_VALUE['indicator'],
+  }),
+  t.type({
+    type: t.literal('textWebContent'),
+    value: ATTACHMENT_VERSION_VALUE['textWebContent'],
+  }),
+]);
+
+type PlanAttachmentData = t.TypeOf<typeof ATTACHMENT_DATA>;
```
Pl217: I think this reference to plan attachment in type name is misleading PlanAttachmentData
```suggestion
type AttachmentData = t.TypeOf<typeof ATTACHMENT_DATA>;
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+
+type PlanAttachmentParent =
```
Pl217: Again, reference to plan attachment is misleading. All attachments are attached to plan, but not all attachments are plan attachments :smile: This type illustrates it well, but its name is not reflecting that
```suggestion
type AttachmentParent =
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+
+type PlanAttachmentResults = Map<AttachmentId, PlanAttachmentResult>;
+
+/**
+ * Get all of the attachments of a particular type for a particular plan.
+ *
+ * In addition to fetching the attachments, this also validates the contents of
+ * their value field, ensuring they are of the correct type.
+ */
+export const getAllPlanAttachments = async ({
```
Pl217: This is a third and last time that I talk about how I think we should not mix attachment with plan attachment and attachment of a plan :) There are lots of references like this throughout this file, I won't comment on each
```suggestion
export const getAllAttachmentsForPlan = async ({
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+export type PlanAttachmentResult = {
+  id: AttachmentId;
+  customRef: string;
+  /**
+   * The part of the plan under which this attachment is directly nested
+   */
+  parent: PlanAttachmentParent;
+  /**
+   * If set, this particular plan attachment is associated with the given
+   * governing entity
```
Pl217: Maybe expand this
```suggestion
   * governing entity. This does not necessarily mean that data available in `parent` is duplicated, because if `type` is `'planEntity'`, this field can be populated via parent of a plan entity
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+  planId,
+  types,
+  planEntities,
+  governingEntities,
+  log,
+}: {
+  database: Database;
+  planId: PlanId;
+  types: AttachmentType[];
+  version: 'latest';
```
Pl217: Why `version`? Not used

Actually, each of these newly added methods have this `version` specified in param types, but not present in params and thus not used

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+   *
+   * Usually all the governing entities for a given plan are provided
+   */
+  governingEntities: MapOfGoverningEntities;
+  log: SharedLogContext;
+}): Promise<PlanAttachmentResults> => {
+  const prototypes = await database.attachmentPrototype.find({
+    where: (builder) => builder.whereIn('type', types),
+  });
+  const attachmentPrototypes = organizeObjectsByUniqueProperty(
```
Pl217: `attachmentPrototypes` is more appropriate for what is called `prototypes`
```suggestion
  const attachmentPrototypesById = organizeObjectsByUniqueProperty(
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+   * A map of governing entities, which **must** include any governing entity
+   * that is the parent of the given attachment.
+   *
+   * Usually all the governing entities for a given plan are provided
+   */
+  governingEntities: MapOfGoverningEntities;
+  log: SharedLogContext;
+}): Promise<PlanAttachmentResults> => {
+  const prototypes = await database.attachmentPrototype.find({
+    where: (builder) => builder.whereIn('type', types),
```
Pl217: Why not filter by `planId` as well?
```suggestion
    where: (builder) => builder.whereIn('type', types).andWhere('planId', planId)
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+  const planAttachments = await database.attachment.find({
+    where: (builder) =>
+      builder.whereIn('type', types).andWhere('planId', planId),
+  });
+  const planAttachmentVersions = await database.attachmentVersion.find({
+    where: (builder) =>
+      builder
+        .whereIn('attachmentId', [
+          ...new Set(planAttachments.map((pa) => pa.id)),
+        ])
```
Pl217: No need to go through `Set`. Attachment IDs are already unique
```suggestion
        .whereIn(
          'attachmentId',
          attachments.map((a) => a.id)
        )
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+    const e = planEntities.get(createBrandedValue(attachment.objectId));
+    if (!e) {
+      throw new Error(
+        `Couldn't find planEntity for attachment ${attachment.id}`
+      );
+    }
+    return { type: 'planEntity', id: e.id };
+  } else {
+    throw new Error(`Invalid objectType for attachment ${attachment.id}`);
+  }
```
Pl217: Could also be without `else`. We had linting rule which we removed in https://github.com/UN-OCHA/hpc-repo-tools/pull/6, but I still prefer reduced nesting :)
```suggestion
  }

  throw new Error(`Invalid objectType for attachment ${attachment.id}`);
```

src/lib/data/attachment.ts
```diff
@@ -0,0 +1,344 @@
...
+  attachment: InstanceDataOfModel<Database['attachment']>;
+  attachmentVersion: InstanceDataOfModel<Database['attachmentVersion']>;
+  log: SharedLogContext;
+}): PlanAttachmentData => {
+  const result = {
+    type: attachment.type,
+    value: attachmentVersion.value,
+  };
+  const decodedValue = ATTACHMENT_DATA.decode(result);
+  if (!isRight(decodedValue)) {
```
Pl217: `isLeft` doesn't get any love :smile:

Pl217 (changes requested):

src/lib/data/planEntities.ts
s0: Yeah i think that was the main motivation I had for using it! :)

src/lib/data/planEntities.ts
s0: Agreed. I'm happy for this to throw an error here when we encounter situations that we don't expect (which will be the behaviour if we encounter bad data).

We could add a `UNIQUE` constraint to the combination of (`childType`, `childId`) fairly easily in a migration which will probably apply without issue, and which would protect us from causing errors at this point?

src/lib/data/planEntities.ts
s0: Nice!

src/lib/data/projects.ts
s0: As far as I can tell, `projectVersionId` is actually not unique in `projectVersionPlan`, though it does form part of a primary key in conjunction with `planId`. I confirmed this by adding two rows with the same `projectVersionId` to a table but differing `planIds`.

That being said, logically a projectVersion should only ever be associated with a single plan, so this schema doesn't make sense as it currently stands, and `projectVersionId` isn't currently duplicated anywhere, so we could add a `UNIQUE` constraint to that column too.

src/lib/data/attachment.ts
s0: See my comment here: https://github.com/UN-OCHA/hpc_service/pull/2439#discussion_r734571242

> I was hoping to make it clear / explicit in the function definition that it's specifically getting the latest versions of everything (and not the current / published versions), and leave room for it being expanded to current later on. Is there an alternative way you think I should do this?

src/lib/data/attachment.ts
s0: haha, you're right reduced nnegation would be good... I typically stick with `isRight` because "right" also means "correct" in this case (i think likely by design haha), but let's switch this out because why not, it's clear enough still.

src/lib/data/governingEntities.ts
s0: Hmm yeah, this does sound messy... Having a bit of a play, It's resulted in me coming up with something like:

```ts
const getRequiredData = <Id, V, P extends string>(
  map: Map<Id, V>,
  entity: {
    [key in P]: Id;
  } & {
    id: string | number;
  },
  property: P,
  mapValueKind = 'object',
  entityKind = 'entity'
): V => {
  const val = map.get(entity[property]);
  if (!val) {
    throw new Error(
      `Missing ${mapValueKind} for ${entityKind} ${entity.id}`
    );
  }
  return val;
}
```

which can be called with something like:

```ts
const prototype = getRequiredData(prototypes, governingEntity, 'entityPrototypeId')
```

Or for the exact same error message:

```ts
const prototype = getRequiredData(prototypes, governingEntity, 'entityPrototypeId', 'prototype', 'governingEntity')
```

If we then had some sort of annotated map, we could get rid of one of those arguments in the call, but that's still a pretty messy function and arguments, which makes me feel it's not worth it to just enforce consistent error messages...

s0: One other thing I've just thought of though. Given that in most cases we're just dealing with database entities, we *could* also annotate / provide some metadata in the entities returned by the database itself. Something hidden when it comes to serializing etc... but still accessible for inspection, which can include the table name, or a general purpose way of representing the object... maybe with some `toString()` implementation that prints out the ID of the entity along with the table... similar to what we pass to `genIdentifier` when validating data upon read... hmm... I'm gonna have a bit of a think...

Pl217 (approved):

PR #35, opened 2021-11-04
Add projectVersionField to v4 models

czmj:


src/db/models/projectVersionField.ts
```diff
@@ -0,0 +1,37 @@
...
+  fields: {
+    generated: {
+      id: { kind: 'branded-integer', brand: PROJECT_VERSION_FIELD_ID },
+    },
+    required: {
+      conditionFieldId: { kind: 'branded-integer', brand: CONDITION_FIELD_ID },
+      projectVersionPlanId: {
+        kind: 'branded-integer',
+        brand: PROJECT_VERSION_PLAN_ID,
+      },
```
Pl217, line 31: `conditionFieldId` and `projectVersionPlanId` are actually `accidentallyOptional` as they aren't defined with `NOT NULL` constraint. 

`required` is for columns with `NON NULL` and `optional` for those without such constraint. But, there are no records with actual `NULL` values for these two columns in the database, qualifying it for `accidentallyOptional`

src/db/models/projectVersionField.ts
```diff
@@ -0,0 +1,37 @@
...
+      value: { kind: 'checked', type: t.any },
```
Pl217: `value` is optional as it has no `NOT NULL` constraint and there are row which have `NULL` as value for this column (unlike `conditionFieldId` and `projectVersionPlanId`) above.

Also, please use `unknown` instead of `any` for json columns.
```suggestion
      value: { kind: 'checked', type: t.unknown },
```

Pl217 (changes requested):

Pl217 (approved):

PR #36, opened 2021-11-04
🎨 Use new db util functions for consistent errors

s0:
Following on from the discussion in https://github.com/UN-OCHA/hpc-api-core/pull/33#discussion_r742675345

Introduce and use new `findAndOrganizeObjectsByUniqueProperty`, `findAndOrganizeObjectsByUniqueValue` and `getRequiredData` functions to consistently check for the existence of certain objects in maps, and fail with an error including relevant identifiable information if they are not found.

This required quite a bit of extra work and a couple of type hacks to get working correctly with as much type-safety as possible.

I spent way too long on this, and ultimately, I'm not convinced that it's worth the additional complexity / abstractions that this adds. It bulks up the util modules with quite a lot of extra cruft, though it does mean the error messages we produce are much more consistent...

Regardless as to whether we want to go with this in the end, it's probably worthwhile at least including 4cfef6f585a0fa4e63c721ad4256b66b86e9961f, which will allow for easier debugging and logging anyway. (Maybe we just keep that and simplify the errors by just including the entity in the error string?

I think before merging this, we should get a couple of other people's opinions on it, if it's too complex, doesn't make sense, let's maybe drop it.

This was a fun little experiment at least 😅

### DIFF
--- a/src/db/fetching.ts
+++ b/src/db/fetching.ts
@@ -1,0 +1,63 @@
+import { AnnotatedMap, organizeObjectsByUniqueValue } from '../util';
+import { Database } from './type';
+import { InstanceDataOfModel, Model } from './util/raw-model';
+import {
+  InstanceOfVersionedModel,
+  VersionedModel,
+} from './util/versioned-model';
+
+/**
+ * Generic type that can be used to obtain the type of an instance of either a
+ * standard table, or a versioned table.
+ */
+type InstanceOf<T extends Database[keyof Database]> = T extends Model<any>
+  ? InstanceDataOfModel<T>
+  : T extends VersionedModel<any, any, any>
+  ? InstanceOfVersionedModel<T>
+  : never;
+
+/**
+ * Fetch a number of items of a particular type from the database,
+ * and organize them into an AnnotatedMap by a unique property.
+ *
+ * The property selected must non-null for every instance retrieved from the
+ * database, otherwise an error will be thrown
+ */
+export const findAndOrganizeObjectsByUniqueProperty = <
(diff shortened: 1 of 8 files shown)


Pl217 (approved):
Lol I first read the code and then the first comment of this PR and I wanted to say almost exact things haha

You will get my approval even though this turned out maybe too specific and other people (currently on the team and future codebase maintainers) will likely find this too hard to grasp.

My thinking is that this all extends methods which are used as a library and you actually don't need to use the methods. One can still use old ways of doing things with fetching data in one call and then using `organizeObjectsByUniqueProperty`, with additional checks later when accessing a map.

On the other hand, even if you don't want to use the methods, you will be forced to read them :smile:
The biggest concern for me is that both of us came to this via evolution, somewhat naturally, and other people will look at usages of these methods and probably have a hard time wrapping their heads around it.

Let's discuss with other team members about what they think. I will not merge because of that reason.

czmj:
@s0 So after reading through the discussion on https://github.com/UN-OCHA/hpc-api-core/pull/33#discussion_r742675345 this totally makes sense and I get why you want to do it. I think on balance it's worth merging although it is quite a lot of code for a small use case - you've added a lot of useful comments and so on to explain it so I don't think it will be too confusing for future people.

The only tiny comment I have on the actual commits is you have a typo in the commit description on f29e5af03b6d1f0ff23f5ea221f5d4981a984758 "Introduce the notion **f** an Annotated Map"

s0:
Cool, fixed commit typo and rebased.

PR #41, opened 2021-11-10
✨ HPC-8260: Required changes to address HPC-8260

s0:
Supports UN-OCHA/hpc_service#2486


czmj:
There's some duplication with #42 in this, but looks good to go once it's rebased.

Pl217 (approved):

PR #42, opened 2021-11-10
Define models needed for HPC-8163

Pl217:
These models are needed for https://github.com/UN-OCHA/hpc_service/pull/2445


czmj (approved):
Thanks!

PR #43, opened 2021-11-11
Accept `null` for `optional` user data

Pl217:
First, `null` should be accepted as user data for `optional` fields. Optional columns are `NULL`-able, so accepting `T | null | undefined` is pretty reasonable.

Other change relates to `accidentallyOptional` in model instances not being treated as nullable, so instead of `T | null`, we will have just `T` (same as old `seqelize` models). Accidentally optional is basically specifying loopholes that never got exploited. The main question is whether we want to validate these fields as nullable as well - [this line](https://github.com/UN-OCHA/hpc-api-core/blob/1c1c8fdeb8adcfe8e33d1757517275255afdce2d/src/db/util/validation.ts#L147).

Without these changes, constructs like this cannot pass type checks:
```ts
const plan = await models.v4.plan.findOne({
  where: { id: createBrandedValue(1) },
});

if (!plan) {
  throw new Error();
}

await models.v4.plan.create({ ...plan });
```

This is needed for https://github.com/UN-OCHA/hpc_service/pull/2445

### DIFF
--- a/src/db/util/model-definition.ts
+++ b/src/db/util/model-definition.ts
@@ -77,5 +77,5 @@ export type UserDataOf<F extends FieldDefinition> = Partial<
   FieldValuesOfSet<F['nonNullWithDefault']>
 > &
   FieldValuesOfSet<F['required']> &
-  Partial<FieldValuesOfSet<F['optional']>> &
-  FieldValuesOfSet<F['accidentallyOptional']>;
+  FieldValuesOfSet<F['accidentallyOptional']> &
+  Partial<Nullable<FieldValuesOfSet<F['optional']>>>;


czmj (dismissed):

s0 (changes requested):
94def0343bae5e7f85a38aaefd90bc0c91d068e3 makes sense to me

The change in e2cc91db959bfa09160dfe4ebf37d3e5efb752ff means that there's effectively no difference between `required` and `accidentallyOptional` as far as the way it's treated in types goes. I'm not neccesarily opposed to this, but then should we just get rid of `accidentallyOptional` entirely, or do we want to still list them diffenently to note that we may want to make schema changes?

Beyond this, if we do make a change to this type, we also need to make a change to the validation to ensure that all accidentallyOptional fields are guaranteed to be set when we read, and throw an error if not. This can be done here: https://github.com/UN-OCHA/hpc-api-core/blob/develop/src/db/util/validation.ts#L147

(it's one of the rare occasions in our codebase where the `io-ts` validation is separate from the types)

Pl217:
> [94def03](https://github.com/UN-OCHA/hpc-api-core/commit/94def0343bae5e7f85a38aaefd90bc0c91d068e3) makes sense to me
> 
> The change in [e2cc91d](https://github.com/UN-OCHA/hpc-api-core/commit/e2cc91db959bfa09160dfe4ebf37d3e5efb752ff) means that there's effectively no difference between `required` and `accidentallyOptional` as far as the way it's treated in types goes. I'm not neccesarily opposed to this, but then should we just get rid of `accidentallyOptional` entirely, or do we want to still list them differently to note that we may want to make schema changes?
> 
> Beyond this, if we do make a change to this type, we also need to make a change to the validation to ensure that all accidentallyOptional fields are guaranteed to be set when we read, and throw an error if not. This can be done here: https://github.com/UN-OCHA/hpc-api-core/blob/develop/src/db/util/validation.ts#L147
> 
> (it's one of the rare occasions in our codebase where the `io-ts` validation is separate from the types)

That is why I committed separately :)

My thinking was that `required` and `accidentallyOptional` are the same, and that `accidentallyOptional` indicates we may want schema changes. After all, that is how I make decisions where to put the column:
- `NOT NULL`: `required`
- `NULL`: Make a query in DB `<column_name> IS NULL` and
  - If there are no records returned, it's `accidentallyOptional`
  - If there are any records, it's `optional`

I have also commented about validation, and included the link to the same line of code, just used future-proof commit hash, instead of `develop` in path :)

If we only keep 94def0343bae5e7f85a38aaefd90bc0c91d068e3, given a little exercise that I used in opening comment
```ts
const plan = await models.v4.plan.findOne({
  where: { id: createBrandedValue(1) },
});

if (!plan) {
  throw new Error();
}

await models.v4.plan.create({ ...plan });
```
that would mean we cannot clone columns marked as accidentaly optional, and would need to provide what developer thinks is a default value. For example, with `plan`, we need to provide value for `restricted`.
```ts
const plan = await models.v4.plan.findOne({
  where: { id: createBrandedValue(1) },
});

if (!plan) {
  throw new Error();
}

await models.v4.plan.create({ ...plan, restricted: false });
```
Providing defaults to `accidentallyOptional` might be fine for boolean and string columns, but what if we want to clone some other model, which has foreign key marked as `accidentallyOptional`, what default value do we provide in such cases.

In [CONTRIBUTING.md](https://github.com/UN-OCHA/hpc-api-core/blob/3b1df698019d677b7ba167dd3b8cf5457a0736d0/CONTRIBUTING.md), there is a section **Special handling of "accidentally optional" fields** where I see that you meant more loose definition of `accidentallyOptional`, so I'll let you decide on this

> The HPC schema has a number fields defined that were accidentally left "nullable" across its tables. Sometimes there is data that is null and sometimes it's just the schema that is incorrect and a migration would fix the issue.
>
> To account for this, models can be defined in a way that allows developers to specify which fields are "accidentally optional". When this is done, the static type-checking requires this field to be specified when creating new rows for a table, but when reading data from the table, it is defined as possibly null.

s0:
Right, I think i envisaged that `accidentallyOptional` would also handle the cases where the majority of the data was there, and the intention of the property was almost certainly for it to be required, but some data has slipped through the cracks.

To be honest... this concept was envisaged at a time where we didn't have any run-time type-checking of the data being returned from the database (and in particular the presence of fields that were optional but probably shouldn't be). So to protect the code, the idea was that we'd mark these fields as possibly null.

Now that we however have runtime type-checking of all data and all fields when read from the database, if we treat  `accidentallyOptional` technically in the same manner as we treat `required`, then we'd at least get runtime validation of these fields, and if we encounter data that is later missing from these fields, we'll get a validation error and not be able to read the data, rather than allowing a value to slip through the type-safe boundary. 

This is still a vast improvement from just assuming possibly null fields are non-null, but it does mean that there's a chance that there is data we wouldn't be able to read. That also being said, `accidentallyOptional` does offer a nice migration path for fields from `optional` to `required`, as it's the only field type that requires a field when creating an entity, but is possibly null when reading. So e.g. you can first prevent any more null values being written, while still being able to read them, and then write a migration using the models, and then later make the field `NON NULL` and change the type to `required`.

So I see our options as some combination of the following things:

1. make `accidentallyOptional` technically equivalent to `required`, just treated as an annotation effectively.
2. get rid of `accidentallyOptional` and just use `required` for these fields. Adding a TODO annotation to those fields with incorrect constraints.
3. rename `accidentallyOptional` to `futureRequired`, or introduce another field type of that name, that requires fields when creating entities, but has the rows potentially null when reading.

I'm also not sure what the best path forward is, maybe (1) and (3)? Having the field as a different type does allow for easy identification of fields that need their constraints changed. We can also add unit-tests that check the constraints of the fields against the definitions at some point too.

Pl217:
Another thing we could do is tackle DB constraints soon. Do we have a ticket to track this work? It should be easy to search for columns marked as `accidentallyOptional` and write a migration for those that have no rows with `NULL` in given column.

I plan to define all (or most) of the models in hpc-api-core soon, so I could also write the migrations, so we can switch these columns to `required`. We can then remove `accidentallyOptional` or treat it as "I don't want this column to be nullable any longer`

s0:
Yeah that sounds good too! I don't think we have a ticket though?

Pl217:
> Yeah that sounds good too! I don't think we have a ticket though?

I created [HPC-8268](https://humanitarian.atlassian.net/browse/HPC-8268) to track this

s0 (approved):

PR #47, opened 2021-11-15
Add createMany function in db library

czmj:
Allow inserting multiple rows at once

### DIFF
--- a/src/db/util/raw-model.ts
+++ b/src/db/util/raw-model.ts
@@ -18,6 +18,13 @@ export type CreateFn<F extends FieldDefinition> = (
   }
 ) => Promise<InstanceDataOf<F>>;
 
+export type CreateManyFn<F extends FieldDefinition> = (
+  data: UserDataOf<F>[],
+  opts?: {
+    trx?: Knex.Transaction<any, any>;
+  }
+) => Promise<Array<InstanceDataOf<F>>>;
+
 export type WhereCond<F extends FieldDefinition> =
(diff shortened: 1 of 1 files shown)


src/db/util/raw-model.ts
```diff
@@ -103,6 +111,12 @@ export const defineRawModel =
       return validateAndFilter(res[0]);
     };
 
+    const createMany: CreateManyFn<F> = async (data, opts) => {
+      const builder = opts?.trx ? tbl().transacting(opts.trx) : tbl();
+      const res = await builder.insert(data).returning('*');
+      return validateAndFilter(res[0]);
```
s0: I'd probably return all of them rather than just the first one (which may also potentially be null / undefined if a zero-length list is given).

src/db/util/raw-model.ts
```diff
@@ -18,6 +18,13 @@ export type CreateFn<F extends FieldDefinition> = (
   }
 ) => Promise<InstanceDataOf<F>>;
 
+export type CreateManyFn<F extends FieldDefinition> = (
+  data: UserDataOf<F>[],
+  opts?: {
+    trx?: Knex.Transaction<any, any>;
+  }
+) => Promise<InstanceDataOf<F>>;
```
s0:
```suggestion
) => Promise<Array<InstanceDataOf<F>>>;
```

s0 (changes requested):

Pl217 (approved):

s0 (approved):

PR #48, opened 2021-11-16
Define `conditionField` and related models

Pl217:
Define `conditionField`, `conditionFieldType` and `conditionFieldReliesOn` models. Will do a release after this is merged.

### DIFF
--- a/src/db/models/conditionField.ts
+++ b/src/db/models/conditionField.ts
@@ -1,7 +1,10 @@
 import * as t from 'io-ts';
 
-import { brandedType } from '../../util/io-ts';
+import { brandedType, EMPTY_TUPLE } from '../../util/io-ts';
 import type { Brand } from '../../util/types';
+import { defineIDModel } from '../util/id-model';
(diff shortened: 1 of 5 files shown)


src/db/models/conditionField.ts
```diff
@@ -12,3 +15,53 @@ export type ConditionFieldId = Brand<
...
+export const CONDITION_FIELD_RULES = t.union([
+  t.partial({
+    conditionality: t.type({ value: t.string }),
+    max: t.number,
+    maxlength: t.number,
+    min: t.number,
+    /**
+     * There is a record in database that has
+     * this typo "mulitple", instead of "multiple"
+     */
```
s0, line 31: Let's add a TODO here and mark this as deprecated too.

s0 (changes requested):
1 minor comment about marking a field as deprecated and adding a TODO comment, otherwise LGTM

s0 (approved):

PR #55, opened 2021-11-25
Expose `DISAGGREGATED_LOCATIONS` codec

krishnareddy-tadi:


src/db/models/json/indicatorsAndCaseloads.ts
```diff
@@ -30,6 +30,28 @@ export type CaseloadOrIndicatorMetricWithValue = t.TypeOf<
...
+    }),
+  ])
+);
+
+export const DISAGGREGATED_LOCATIONS_WITH_PCODE = t.array(
+  t.intersection([
+    DISAGGREGATED_LOCATIONS.type,
+    t.partial({ pcode: t.union([t.string, t.number, t.undefined, t.null]) }),
+  ])
+);
```
Pl217: Don't define `DISAGGREGATED_LOCATIONS_WITH_PCODE` since you need this codec as a specific case, describing data returned from your endpoint.

Instead, extract `DISAGGREGATED_LOCATIONS` to a separate variable, `export` it and define `DISAGGREGATED_LOCATIONS_WITH_PCODE` in the file you use it.

Pl217 (changes requested):

Pl217 (approved):

PR #57, opened 2021-11-25
✨ Introduce getConditionFieldsForProjects

s0:
This supports HPC-6264

(this will cause type-error conflict with #66, so whichever is merged last will need to be rebased and fixed)

### DIFF
--- a/src/lib/data/projects.ts
+++ b/src/lib/data/projects.ts
@@ -11,6 +11,7 @@ import { Op } from '../../db/util/conditions';
 export interface ProjectData {
   project: InstanceOfModel<Database['project']>;
   projectVersion: InstanceOfModel<Database['projectVersion']>;
+  projectVersionPlan: InstanceOfModel<Database['projectVersionPlan']>;
   workflowStatus: InstanceOfModel<Database['workflowStatusOption']> | null;
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/data/projects.ts
```diff
@@ -167,3 +169,38 @@ export async function getGoverningEntityIDsForProjects({
 
   return result;
 }
+
+export const getConditionFieldsForProjects = async ({
+  database,
+  projects,
+}: {
+  database: Database;
+  projects: Map<ProjectId, ProjectData>;
```
Pl217: What do you think about making this method easier to use outside your use-case, by requiring less project data?
```suggestion
  projects: Map<ProjectId, Pick<ProjectData, 'projectVersionPlan'>>;
```

Pl217 (commented):
It makes sense to make these changes in two separate commits, because changes are pretty different. In first commit, extend `getAllProjectsForPlan` and in second commit introduce `getConditionFieldsForProjects`

s0:
Rebased ontop of `develop` now #66 is merged, this is ready for review now.

No need to bump the version with this one, as there is a bit more work that is still needed for UN-OCHA/hpc_service#2514

Pl217 (approved):

PR #58, opened 2021-11-26
Define budget segment models

Pl217:
Commits cherry picked from #46. These models are needed for https://github.com/UN-OCHA/hpc_service/pull/2504

### DIFF
--- a/src/db/models/budgetSegmentBreakdown.ts
+++ b/src/db/models/budgetSegmentBreakdown.ts
@@ -2,6 +2,8 @@ import * as t from 'io-ts';
 
 import { brandedType } from '../../util/io-ts';
 import type { Brand } from '../../util/types';
+import { defineIDModel } from '../util/id-model';
+import { BUDGET_SEGMENT_ID } from './budgetSegment';
 
 export type BudgetSegmentBreakdownId = Brand<
   number,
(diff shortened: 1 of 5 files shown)


src/db/models/budgetSegmentBreakdown.ts
```diff
@@ -13,3 +15,61 @@ export const BUDGET_SEGMENT_BREAKDOWN_ID = brandedType<
...
+          amount: t.number,
+          percent: t.number,
+          description: t.string,
+        }),
+        t.partial({
+          fixed: BUDGET_SEGMENT_BREAKDOWN_FIXED_TYPE,
+        }),
+      ])
+    ),
+  }),
```
rgautam98, line 55: I need a second opinion on this part from @s0 otherwise, everything looks good.

rgautam98 (approved):

src/db/models/budgetSegmentBreakdown.ts
Pl217: These were extracted based on existing data that we already have in the database. Since we do v[alidation when reading records](https://github.com/UN-OCHA/hpc-api-core/blob/33e445cbf8794bb062f18ee65fd6b43e505538ee/src/db/util/raw-model.ts#L123), I defined everything that budget segment breakdown content JSON data can have.

Now, since JSON data was not validated in the past, there is garbage data in many cases, but I tried capturing all possibilities for the sake of reading from DB. There is an idea to split into the codecs for reading existing data and codecs creating new records, where we want to be very strict going forward.

PR #26, opened 2021-10-31
HPC-8232, HPC-7811, HPC-8130 - setup testing environment with jest and docker

ashwinkjoseph:
Set up unit testing using Jest and Docker. In Docker compose, empty Postgres database is initialized using a minimal schema that matches what’s currently being used in HPC. The schema was generated from Postgres container from [hpc-suite](https://github.com/UN-OCHA/hpc-suite)'s `start.sh` with latest production DB synced to local:
```shell
docker exec hpc_service-pgsql-1 pg_dump --schema-only --no-owner -U postgres hpc --file /backups/schema-2024-02-29.sql
```

A few example tests were added and those which interact with models are using transactions.

This PR is related to tickets [HPC-8232](https://humanitarian.atlassian.net/browse/HPC-8232), [HPC-7811](https://humanitarian.atlassian.net/browse/HPC-7811) and [HPC-8130](https://humanitarian.atlassian.net/browse/HPC-8130).

[HPC-8232]: https://humanitarian.atlassian.net/browse/HPC-8232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HPC-7811]: https://humanitarian.atlassian.net/browse/HPC-7811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HPC-8130]: https://humanitarian.atlassian.net/browse/HPC-8130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #66, opened 2021-12-02
✨ Introduce new custom where condition construction 

s0:
As discussed earlier today, this PR introduces a new abstraction over where conditions (with support for nested conjunctions (AND) and disjunctions (OR)), to allow for better type-safety with data used for constructing queries, and to make our most common use-cases easier to write.

The queries generated by this are unit-tested in hpc_service: UN-OCHA/hpc_service#2529. (This is needed until #26 is merged)

This PR makes use of query builders in the where clauses type errors, using query builders in conditions is still possible, but now requires us to explicitly use `[Cond.BUILDER]`. This should help us catch and rewrite all instances of usage of the old builder when not neccesary.

(this will cause type-error conflict with #57, so whichever is merged last will need to be rebased and fixed)

### DIFF
--- a/src/db/util/conditions.ts
+++ b/src/db/util/conditions.ts
@@ -6,7 +6,7 @@ import Knex = require('knex');
 /**
  * Symbols to use for where condition construction
  */
-namespace ConditionSymbols {
+export namespace ConditionSymbols {
   export const BUILDER = Symbol('builder');
   export const AND = Symbol('and');
   export const OR = Symbol('or');
@@ -26,7 +26,7 @@ export const Cond = ConditionSymbols.Cond;
 /**
  * Symbols to use when constructing conditions for a single property
  */
-namespace PropertySymbols {
+export namespace PropertySymbols {
   export const IN = Symbol('in');
   export const NOT_IN = Symbol('not in');
 
--- a/src/db/util/conditions.ts
+++ b/src/db/util/conditions.ts
@@ -44,10 +44,10 @@ export const Op = PropertySymbols.Op;
 namespace PropertyConditions {
   export type EqualityCondition<T> = T;
   export type InCondition<T> = {
-    [Op.IN]: T[];
+    [Op.IN]: Iterable<T>;
   };
   export type NotInCondition<T> = {
-    [Op.NOT_IN]: T[];
+    [Op.NOT_IN]: Iterable<T>;
   };
   /**
    * A condition that must hold over a single property whose type is T
@@ -167,9 +167,9 @@ export const prepareCondition =
         if (PropertyConditions.isEqualityCondition(propertyCondition)) {
           builder.where(property, propertyCondition);
         } else if (PropertyConditions.isInCondition(propertyCondition)) {
-          builder.whereIn(property, propertyCondition[Op.IN]);
+          builder.whereIn(property, [...propertyCondition[Op.IN]]);
         } else if (PropertyConditions.isNotInCondition(propertyCondition)) {
-          builder.whereNotIn(property, propertyCondition[Op.NOT_IN]);
+          builder.whereNotIn(property, [...propertyCondition[Op.NOT_IN]]);
         } else {
           throw new Error(`Unexpected condition type: ${propertyCondition}`);
         }
--- a/src/db/util/conditions.ts
+++ b/src/db/util/conditions.ts
@@ -260,7 +260,9 @@ export const prepareCondition =
         PropertyConditions.Condition<InstanceData[keyof InstanceData]>
       ][];
       for (const [property, propertyCondition] of propertyConditions) {
-        if (PropertyConditions.isEqualityCondition(propertyCondition)) {
+        if (propertyCondition === undefined) {
+          throw new Error(`Unexpected undefined value for ${property}`);
+        } else if (PropertyConditions.isEqualityCondition(propertyCondition)) {
           builder.where(property, propertyCondition);
         } else if (PropertyConditions.isInCondition(propertyCondition)) {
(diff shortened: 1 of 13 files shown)


src/db/util/conditions.ts
```diff
@@ -0,0 +1,178 @@
...
+// we use namespaces here to neatly organize and scope the many type and type guard definitions
+
+import Knex = require('knex');
+
+/**
+ * Symbols to use for where condition construction
+ */
+export namespace ConditionSymbols {
+  export const BUILDER = Symbol('builder');
+  export const AND = Symbol('and');
```
Pl217, line 11: What exactly would be the benefit of using `AND`, besides completeness with having `OR`? When properties are supplied to `where`, `AND` is the default. Is there some construct where this would be useful, that I am not seeing?

If not, I would like to drop `AND` because I would not like to keep telling people during reviews: "Hey, don't use `AND` since it is a default behavior anyway, let's simplify this`

src/db/util/conditions.ts
```diff
@@ -0,0 +1,178 @@
...
+   */
+  export type Condition<T> =
+    | EqualityCondition<T>
+    | InCondition<T>
+    | NotInCondition<T>;
+
+  export const isEqualityCondition = <T>(
+    condition: Condition<T>
+  ): condition is EqualityCondition<T> =>
+    typeof condition !== 'object' || condition === null;
```
Pl217, line 117: I wanted to break this by supplying `undefined` (literal) and see what happens. Since the condition resolves to `true`, `undefined` is supplied to `knex`'s `where` and we get "Undefined binding(s) detected when compiling WHERE" error.

I don't know what to make out of this though. I don't think sequelize failed in this case (not sure), but I don't want to silently convert it to `null` check either. No one will supply `undefined` literal (hopefully), but maybe it comes from a variable. Any thoughts on this?

src/db/util/conditions.ts
```diff
@@ -0,0 +1,178 @@
...
+export namespace PropertySymbols {
+  export const IN = Symbol('in');
+  export const NOT_IN = Symbol('not in');
+
+  /**
+   * Symbols to use when constructing conditions for a single property
+   */
+  export const Op = {
+    IN: IN,
+    NOT_IN: NOT_IN,
```
Pl217, line 46: Great stuff! I would also like us to be able to do `IS NOT NULL` check and `LIKE` without using a builder. At least those two, an maybe `LT`, `LTE`, `GT` and `GTE` should be added, mainly for `date` type comparison.

Pl217 (changes requested):
Another brilliant piece! :clap:

My inline comments are mostly on expanding this, because I would like some more operators added.

src/db/util/conditions.ts
s0: Yeah, in cases where we want to use OR or a query builder in one of the AND clauses, or if we want to combine some query builder stuff with some normal property stuff. Something like:

`(A or B) and (B or C)` or `(A or B) and C`.

something like this in the unit tests: https://github.com/UN-OCHA/hpc_service/blob/1a0166ffb0d5fe47934de8a5cf6b51566a65f8c5/test/hpc-api-core/src/db/util/conditions.spec.ts#L129-L136

Or:

```ts
{
  [Cond.AND]: [
    {
      id: 123,
    },
    {
      [Cond.BUILDER]: (qb) => qb.whereNotNull('id'),
    },
  ],
}
```

But yeah, in the cases where we're ONLY using AND, and not using an OR or BUILDER inside of an AND clause, we'd want to use the simpler / cleaner construction.

src/db/util/conditions.ts
s0: I was thinking about that too, unsure whether it was worth adding before we actually had a use, but it won't be too much extra work :)

src/db/util/conditions.ts
s0: I dont think this is something that we can easuly make a type error due to the "Partial" nature of the condition over properties, as long as something is defined as `[...]?: T`, then i think you can assign undefined to it. If that's the case, then I think we probably just want to throw an error. If we're not happy with it being given to `knex` to catch, we can try and catch it earlier, which would happen if we just modify this condition slightly

src/db/util/conditions.ts
Pl217: Ok, I understand. I didn't realize immediately that
```ts
if (OverallConditions.isQueryBuilderCondition(condition)) {
  builder.where(condition[Cond.BUILDER]);
}
```
swallows "plain" conditions, so I thought you could write the same thing as:
```ts
{
  id: 123,
  [models.v4.Cond.BUILDER]: (qb) => qb.whereNotNull('id'),
}
```

src/db/util/conditions.ts
```diff
@@ -0,0 +1,303 @@
...
+
+  export type InCondition<T> = {
+    [Op.IN]: Iterable<T>;
+  };
+  export type NotInCondition<T> = {
+    [Op.NOT_IN]: Iterable<T>;
+  };
+
+  export type IsNullCondition = {
+    [Op.IS_NULL]: boolean;
```
Pl217, line 71: I was thinking that we should use `{propertyName: null}` for `propertyName IS NULL` and introduce `[Op.IS_NOT_NULL]: null` or `[Op.NE]: null`, but I am fine with what you used.

It is saying that `NULL` should be treated as a special case, and even though there are two ways to write `IS NULL` now (`{propertyName: null}` is more likely to be used), let's not complicate any further and use what you suggested.

Pl217 (approved):

PR #67, opened 2021-12-02
✨ HPC-8267: Introduce getProjectBudgetsByOrgAndCluster

s0:
Supports https://github.com/UN-OCHA/hpc_service/pull/2514

This is an up-to-date implementation of getProjectBudgetsByOrg from
hpc_service, which uses all the latest coding practices for the project.

Once this has been merged and is usable, getProjectBudgetsByOrg will be updated in https://github.com/UN-OCHA/hpc_service/pull/2514 to be implemented using this new function.

### DIFF
--- a/src/lib/data/projects.ts
+++ b/src/lib/data/projects.ts
@@ -7,6 +7,16 @@ import { Database } from '../../db/type';
 import { InstanceOfModel } from '../../db/util/types';
 import { getRequiredData, groupObjectsByProperty, isDefined } from '../../util';
 import { Op } from '../../db/util/conditions';
+import { GlobalClusterId } from '../../db/models/globalCluster';
+import { createBrandedValue } from '../../util/types';
+import { SharedLogContext } from '../logging';
+import { isEqual } from 'lodash';
+
+/**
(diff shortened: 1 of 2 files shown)


src/lib/data/projects.ts
```diff
@@ -228,3 +238,216 @@ export const getConditionFieldsForProjects = async <
...
+    'projectVersionId'
+  );
+
+  for (const [projectId, p] of projects.entries()) {
+    // There should be only 1 segment with name BUDGET_SEGMENTATION_BY_ORG
+    // for any project (except those with no budget)
+    const segments = segmentsByProjectVersion.get(p.projectVersion.id);
+
+    // Handle $0 projects with no budget segments
+    if (!segments && p.projectVersion.currentRequestedFunds === '0') {
```
Pl217, line 330: Too bad we aren't ready to use `BigInt` here and also below where you do `if (sum.toString() !== p.projectVersion.currentRequestedFunds)`. We cannot even convert `p.projectVersion.currentRequestedFunds` into `BigInt`, because we cannot compare to `0n` literal, since our compilation target is lower than ES2020.

Pl217 (approved):

PR #80, opened 2022-02-02
Remove references to hidId

czmj:


Pl217 (approved):
I'm approving this, but let's not merge until needed

src/auth/index.ts
```diff
@@ -168,39 +157,30 @@ export const getLoggedInParticipant = async (
...
+
+    await activateInvitesForEmail(
+      participant,
+      hidInfo.email,
+      context,
+      processInvite
+    );
+  }
+
+  if (!participant.hidSub) {
```
Pl217: This check is not needed at all.

You replace `hidId` with `hidSub`, but in fact, if we didn't have `hidInfo.sub` above, we would exit this method early.
If we did have a value for `hidInfo.sub`, we would either find a participant by this value or create a new one, so this check is useless.

Pl217 (changes requested):

Pl217 (approved):

PR #82, opened 2022-03-08
HPC-8397: Define excel download job type

krishnareddy-tadi:


src/db/models/job.ts
```diff
@@ -38,13 +40,23 @@ export const PDF_FILE_RECORD = t.type({
...
 export const JOB_METADATA_PROJECT_PDF = t.partial({
   startedBy: PARTICIPANT_ID,
   processed: t.number,
   total: t.number,
   files: t.array(PDF_FILE_RECORD),
 });
 
+export const JOB_METADATA_PROJECT_EXCEL = t.partial({
+  startedBy: PARTICIPANT_ID,
+  files: t.array(EXCEL_FILE_RECORD),
```
Pl217: Why have an array of files if you're only going to store one file?

Pl217 (changes requested):

src/db/models/job.ts
```diff
@@ -38,13 +40,28 @@ export const PDF_FILE_RECORD = t.type({
...
 export const JOB_METADATA_PROJECT_PDF = t.partial({
   startedBy: PARTICIPANT_ID,
   processed: t.number,
   total: t.number,
   files: t.array(PDF_FILE_RECORD),
 });
 
+export const JOB_METADATA_PROJECT_EXCEL = t.partial({
+  startedBy: PARTICIPANT_ID,
+  files: EXCEL_FILE_RECORD,
```
Pl217: If it's no longer an array, this should be `file`, not `files`
```suggestion
  file: EXCEL_FILE_RECORD,
```

Pl217 (changes requested):

src/db/models/job.ts
```diff
@@ -38,13 +40,28 @@ export const PDF_FILE_RECORD = t.type({
   projectId: PROJECT_ID,
 });
 
+export const EXCEL_FILE_RECORD = t.type({
+  name: t.string,
+  planId: PLAN_ID,
```
Pl217: Why include `planId` here? I think you modeled this codec after `PDF_FILE_RECORD` (above), but we don't need the same information. Also, we cannot download Excel files in bulk, like we can do for PDF.

Since you only need file name, I suggest you drop this codec and add `fileName: t.string` to `JOB_METADATA_PROJECT_EXCE`

Pl217 (approved):

PR #83, opened 2022-03-11
HPC-8404: Disaggregation models created by cluster leads are not visible to other in the same cluster- it should be

krishnareddy-tadi:


Pl217:
The goal is to remove `creator` column from `disaggregationModel` table. That `creator` column stores JSON data where keys are `participantHidId` and `roles`.

We plan to migrate `participantHidId` field to a separate column in https://github.com/UN-OCHA/hpc_service/pull/2654/. Instead of making `roles` optional, let's keep it and remove entire `creator` column later.

In the transitory period, let's keep `roles` field. We can set it to empty array, but we can also keep old behavior. Since we're going to remove `creator` column, both are valid options.

PR #88, opened 2022-03-23
HPC-8409:Temporary fix to facilitate Anonymisation of Orgs

krishnareddy-tadi:


src/db/models/flowObjectType.ts
```diff
@@ -16,6 +16,7 @@ export const FLOW_OBJECT_TYPE_TYPE = t.keyof({
   planEntity: null,
   project: null,
   usageYear: null,
+  anonymisedOrganization: null,
```
Pl217: Let's keep alphabetical order already established. Put this value at the top

Pl217 (changes requested):
You will need to create a migration to insert a new row in `flowObjectType` table. Also, you have a typo in your first commit message :)

Pl217 (approved):

PR #95, opened 2022-05-06
HPC-8382: Update `node-fetch` to `v3`

Pl217:
This PR will need to wait until problems with `jest` are resolved - https://github.com/node-fetch/node-fetch/issues/1289 Until then, we cannot run tests in hpc_service


Pl217:
We are removing `node-fetch` dependency in HPC-10144

PR #97, opened 2022-05-18
HPC-8026:Allow selecting monitoring periods to include in plan publishing

krishnareddy-tadi:


src/db/models/planTag.ts
```diff
@@ -36,6 +36,11 @@ export default defineIDModel({
     optional: {
       comment: { kind: 'checked', type: t.string },
       revisionState: { kind: 'checked', type: PLAN_TAG_REVISION_STATE },
+      type: { kind: 'checked', type: t.string },
```
Pl217: This column looks like a candidate for either enum or a limited set of string values. If you go with enum route, you will need to add the enum in DB migration. `flowObject` is a [good example](https://github.com/UN-OCHA/hpc-api-core/blob/31e17734fc303f1f9fbf4e2005b70a94eb7c47ee/src/db/models/flowObject.ts#L32) that employs both ways, `behavior` column is an enum, whereas `refDirection` and `objectType` have a set of allowed strings.

src/db/models/planTag.ts
```diff
@@ -36,6 +36,11 @@ export default defineIDModel({
...
+      reportingPeriods: {
```
Pl217: Given that you defined `reportingPeriods` as array of integers in [the migration](https://github.com/UN-OCHA/hpc_service/pull/2744/files#diff-e42de6ac05965509da24c368226d605673b54424d12dc42b219265ddeab55ef8R8), which defaults to an empty string, this column does not belong in `optional`, but in `nonNullWithDefault`.

Pl217 (changes requested):

src/db/models/planTag.ts
```diff
@@ -5,6 +5,11 @@ import type { Brand } from '../../util/types';
 import { defineIDModel } from '../util/id-model';
 import { PLAN_ID } from './plan';
 
+const PLAN_PUBLISH_TYPE = t.keyof({
+  Major: null,
+  Minor: null,
+  Custom: null,
+});
```
Pl217: I don't think any other similar types are capitalized.
```suggestion
const PLAN_PUBLISH_TYPE = t.keyof({
  major: null,
  minor: null,
  custom: null,
});
```

Pl217 (changes requested):

Pl217 (approved):

PR #98, opened 2022-06-08
HPC-8483: Add `'organizationLevel'` category group

Pl217:
This PR is needed for https://github.com/UN-OCHA/hpc_service/pull/2757

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unocha/hpc-api-core",
-  "version": "4.5.0",
+  "version": "4.5.1",
   "description": "Core libraries supporting HPC.Tools API Backend",
   "license": "Apache-2.0",
   "private": false,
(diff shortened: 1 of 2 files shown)


krishnareddy-tadi (approved):

Pl217:
Me and Nick have decided to change the approach here, so that we won't introduce any new columns to `category` table. With structure "OrgType --(1 to Many)--> OrgSubType <--(Many to 1)-- OrgLevel" in mind (from HPC-8483), we see that only OrgLevel parent needs to be added to OrgSubType, since OrgType parent is already attributed via `parentID` column.

I wanted to add `levelParentId` column to the `category` table initially, but Nick proposed to add another "virtual" entity called OrgSubLevel. Instead of matching OrgLevel parent via ID parameter to OrgSubType, they will be matched by name between OrgSubLevel and OrgSubType. When looking up OrgLevel, we need to find category with `group = 'organizationLevel'` and same name as OrgSubType and then use `parentID` column from the result to find the actual OrgLevel.

Essentially, rows will be used as foreign keys references, not columns. This is the design Nick asked for and I don't yet see any problems, besides this not feeling natural to relation databases. The premise for this is that names of OrgType/OrgSubType... cannot change from the UI, introducing potential bugs, but are only accessible to developer scripts and DB migrations.

PR #124, opened 2023-08-30
HPC-9107: Allow cluster leads to grant ownership of project

enxtur:


src/auth/permissions.ts
```diff
@@ -203,6 +203,11 @@ export const AUTH_PERMISSIONS = {
      * Can move projects associated with this cluster to any step of the workflow.
      */
     PROJECT_WORKFLOW_MOVE_IF_PLAN_UNLOCKED: 'projectWorkflowMoveIfPlanUnlocked',
+    /**
+     * Can edit access and permissions of all projects.
```
Pl217: This gives off impression that this permission is global, for ALL projects. Let's be more specific
```suggestion
     * Can edit access and permissions of all projects for a given governing entity.
```

Pl217 (changes requested):

Pl217 (approved):

PR #126, opened 2023-09-14
HPC-9189: Remove PLAN_TAG_REVISION_STATE from planTag

manelcecs:
Remove duplicated plan revision state codecs from plan and planTag models

### DIFF
--- a/src/db/models/planTag.ts
+++ b/src/db/models/planTag.ts
@@ -3,7 +3,7 @@ import * as t from 'io-ts';
 import { brandedType } from '../../util/io-ts';
 import type { Brand } from '../../util/types';
 import { defineIDModel } from '../util/id-model';
-import { PLAN_ID } from './plan';
+import { PLAN_ID, PLAN_REVISION_STATE } from './plan';
 
 export type PlanTagId = Brand<
   number,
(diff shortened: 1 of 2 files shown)


src/db/models/planTag.ts
```diff
@@ -4,6 +4,7 @@ import { brandedType } from '../../util/io-ts';
 import type { Brand } from '../../util/types';
 import { defineIDModel } from '../util/id-model';
 import { PLAN_ID } from './plan';
+import { PLAN_REVISION_STATE } from './plan';
```
Pl217: You import from same file on two different lines, you should combine them into:
```ts
import { PLAN_ID, PLAN_REVISION_STATE } from './plan';
```

Pl217 (changes requested):
One minor thing inline. I will fix it myself. Regarding what you wrote in the first comment:
> Remove duplicated codecs from plan and planTag models in hpc-api-core.
> 
> Once this is merge, we need to update both HPC_service and HPC-api

* We are in hpc-api-core repo, so no need to be specific about this :)
* We don't need to update hpc_service and hpc-api to include the latest version because they can keep functioning with how things were. The only reason why we would potentially need to update is if any of them was specifically importing `PLAN_TAG_REVISION_STATE`, which we now removed. Therefore, I would say these changes can wait and need not be released immediately, but we can bundle them with some other change which affects hpc_service/hpc-api directly and release at that time.

Pl217 (approved):

PR #128, opened 2023-09-25
Add a bunch of utility methods supporting work on HPC-9193

Pl217:
Introduced utility methods to get:
- Years of a plan
- Basic organization info plus their type, subtype and level
- Attachment and measurement disaggregation data
- All country locations

Also, some non-fetching utility methods were added to:
- Convert string to camel case
- Clean numeric values; `cleanNumberVal` got moved here from hpc_service

### DIFF
--- a/src/lib/data/attachments.ts
+++ b/src/lib/data/attachments.ts
@@ -8,6 +8,8 @@ import {
 } from '../../db/models/attachmentPrototype';
 import { GoverningEntityId } from '../../db/models/governingEntity';
 import { ATTACHMENT_VERSION_VALUE } from '../../db/models/json/attachment';
+import type { CaseloadOrIndicatorMetricsValues } from '../../db/models/json/indicatorsAndCaseloads';
+import { LOCATION_ID } from '../../db/models/location';
 import { PlanId } from '../../db/models/plan';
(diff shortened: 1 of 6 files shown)


src/lib/data/attachments.ts
```diff
@@ -350,3 +356,270 @@ export const composeCustomReferenceForAttachment = ({
   customReference += `${prototype.refCode}${attachmentVersion.customReference}`;
   return customReference;
 };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const disaggregationData = <T extends t.Type<any>>(additionalFields: T) =>
```
Delgee: We can extend `t.Mixed` instead of `t.Type<any>`

Delgee (approved):

src/lib/data/attachments.ts
Pl217: Thanks, very good suggestion!

I've looked into io-ts type definitions and found:
```ts
export interface Any extends Type<any, any, any> {}
export interface Mixed extends Type<any, any, unknown> {}
export declare class AnyType extends Type<any> {...}
```

According to this, I could use `AnyType` to replace what I have used and get rid of linting warning (without suppressing it), but I'm gonna use what you suggested.

PR #132, opened 2023-10-30
HPC-8569: Add `legacy` model

manelcecs:
This PR add the model 'legacy' in order to achieve feature parity with fts-admin v1

### DIFF
--- a/src/db/models/legacy.ts
+++ b/src/db/models/legacy.ts
@@ -1,0 +1,16 @@
+import * as t from 'io-ts';
+
+import { defineSequelizeModel } from '../util/sequelize-model';
+import { FLOW_OBJECT_TYPE_TYPE } from './flowObjectType';
+
+export default defineSequelizeModel({
+  tableName: 'legacy',
+  fields: {
+    required: {
+      objectType: { kind: 'checked', type: FLOW_OBJECT_TYPE_TYPE },
+      objectID: { kind: 'checked', type: t.number },
+      legacyID: { kind: 'checked', type: t.number },
+    },
+  },
+  softDeletionEnabled: false,
+});
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -52,6 +52,7 @@ import iatiRecipientCountry from './models/iatiRecipientCountry';
 import iatiTransaction from './models/iatiTransaction';
 import job from './models/job';
 import jobAssociation from './models/jobAssociation';
+import legacy from './models/legacy';
 import location from './models/location';
 import lookup from './models/lookup';
 import measurement from './models/measurement';
@@ -164,6 +165,7 @@ const initializeTables = (conn: Knex) => ({
   iatiTransaction: iatiTransaction(conn),
   job: job(conn),
   jobAssociation: jobAssociation(conn),
+  legacy: legacy(conn),
   location: location(conn),
   lookup: lookup(conn),
   measurement: measurement(conn),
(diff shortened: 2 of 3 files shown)


src/db/util/raw-model.ts
```diff
@@ -181,6 +195,15 @@ export const defineRawModel =
...
     };
 
+    const count: CountFn<F> = async ({ where, trx } = {}) => {
+      const builder = trx ? tbl().transacting(trx) : tbl();
+      const query = builder.where(prepareCondition(where ?? {})).count('*');
+
+      const res = await query;
+
+      return res.map((val) => val as unknown);
+    };
```
Pl217: The existence of `count` method isn't really justified. It doesn't provide any value. You can do the exact same thing by using `find` and then checking `length` property of results.

The mapping `res.map((val) => val as unknown)`, which you do at the end of this method isn't doing anything actually. You are just looping through an array and returning its elements.

src/db/util/raw-model.ts
```diff
@@ -172,6 +182,10 @@ export const defineRawModel =
         query.orderBy(orderBy);
       }
 
+      if (orderByRaw !== undefined && orderByRaw !== '') {
+        query.orderByRaw(orderByRaw);
+      }
```
Pl217: For raw queries, you should use `queryBuilder`, like you are doing in test file `tests/db/conditions.spec.ts`. Just remove `orderByRaw` option. When rebasing, just drop commits ef7d6ee6008cb89e4e93a42c9cd9e09bbc7281c4 and 53a68a7e4d6a21699cfcc962727586310dd34f18.

src/db/models/legacy.ts
```diff
@@ -0,0 +1,15 @@
...
+  tableName: 'legacy',
+  fields: {
+    required: {
+      objectType: { kind: 'checked', type: t.string },
+      objectID: { kind: 'checked', type: t.number },
+      legacyID: { kind: 'checked', type: t.number },
+    },
+  },
+  softDeletionEnabled: false,
+});
```
Pl217: Model for this table wasn't defined on purpose. As its name suggests, this table is storing legacy data, which we no longer want. I see that newest record of this table is from January 2019, which is more than 5 years old.

On the other hand, I see that this model is used in https://github.com/UN-OCHA/hpc-api/pull/219, but I am not sure whether we truly need it or not. Maybe we need to use it for the sake of completeness of the API, to support old flows. If you are sure we need to use it, then keep this model, but change the type of `objectType` to `FLOW_OBJECT_TYPE_TYPE`.
```suggestion
export default defineSequelizeModel({
  tableName: 'legacy',
  fields: {
    required: {
      objectType: { kind: 'checked', type: FLOW_OBJECT_TYPE_TYPE },
      objectID: { kind: 'checked', type: t.number },
      legacyID: { kind: 'checked', type: t.number },
    },
  },
  softDeletionEnabled: false,
});
```

Pl217 (changes requested):

src/db/util/raw-model.ts
manelcecs: After different changes through development I agree with you that this is not longer needed.
The intention of doing this was to reduce execution times when performing large search queries while pagination.
Doing this the execution time was reduced notably. 
Nonetheless This is no longer needed - I'll remove it.

src/db/models/legacy.ts
manelcecs: As peer as your comment this is needed to provide full compatibility with the previous version of fts-admin.
fts-admin v2 must provide pair-functionality with fts-admin.

Pl217 (approved):

PR #137, opened 2024-01-24
HPC-9354: Add `visibilityPreferences` to `planVersion` model

enxtur:
Added disaggregation visibility fields to `planVersion` model.

### DIFF
--- a/src/db/models/planVersion.ts
+++ b/src/db/models/planVersion.ts
@@ -20,6 +20,11 @@ const PLAN_VERSION_CLUSTER_SELECTION_TYPE = t.keyof({
   multi: null,
 });
 
+const PLAN_VISIBILITY_PREFERENCES = t.type({
+  isDisaggregationForCaseloads: t.boolean,
+  isDisaggregationForIndicators: t.boolean,
+});
+
 export default defineLegacyVersionedModel({
   tableName: 'planVersion',
   fields: {
@@ -28,6 +33,10 @@ export default defineLegacyVersionedModel({
     },
     nonNullWithDefault: {
       isForHPCProjects: { kind: 'checked', type: t.boolean },
+      visibilityPreferences: {
+        kind: 'checked',
+        type: PLAN_VISIBILITY_PREFERENCES,
+      },
     },
     required: {
       planId: { kind: 'branded-integer', brand: PLAN_ID },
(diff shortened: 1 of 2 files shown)


src/auth/permissions.ts
```diff
@@ -91,6 +91,11 @@ export const AUTH_PERMISSIONS = {
      */
     CHANGE_ANY_PLAN_VISIBILITY_IN_PROJECTS:
       'canChangeAnyPlanVisiblityInProjects',
+    /**
+     * Can change disaggregation visibility of any plan in Projects Module
+     */
+    CHANGE_ANY_PLAN_DISAGGREGATION_VISIBILITY_IN_PROJECTS:
+      'canChangeAnyPlanDisaggregationVisiblityInProjects',
```
Pl217: I don't think there is a need for new permissions. Not because they are granted to same roles, because we shouldn't make permissions reusable just because they are granted to same roles.

Rather, I think we should unify these visibility changing settings under one page, and then we can control access with same permission.

Honestly, I think this feature to hide disaggregations is not useful at all and is just wasting our time, and I [commented about that](https://humanitarian.atlassian.net/browse/HPC-9354?focusedCommentId=168246) on the ticket.

src/db/models/planVersion.ts
```diff
@@ -28,6 +28,11 @@ export default defineLegacyVersionedModel({
     },
     nonNullWithDefault: {
       isForHPCProjects: { kind: 'checked', type: t.boolean },
+      isDisaggregationVisibleForCaseloads: { kind: 'checked', type: t.boolean },
+      isDisaggregationVisibleForIndicators: {
+        kind: 'checked',
+        type: t.boolean,
+      },
```
Pl217: In similar fashion to my previous comment, I don't want to add new columns to the database, just because someone wanted to hide something in the UI.

The names are huge, but I cannot come up with a shorter name, because disaggregation, visible/visibility and caseload/indicator are all long words on their own.

This feature to hide disaggregations is completely unnecessary in my opinion, but clients may want to hide/show more stuff in the future, thus, I want to prevent the need to add more columns with long names in the future.

So, I would combine all three (2 new columns and previously existing `isForHPCProjects`) into a single JSON column. This JSON column would need to be controlled by `io-ts` codec, to prevent bad input. This will honestly complicate things a bit and need some refactoring, so whether we want to do it depends on the discussion that will happen on the ticket, since I want to unify these visibility options under same setting page.

Pl217 (changes requested):

Pl217 (approved):

PR #142, opened 2024-03-21
Fix test shell script

Pl217:
Also, use newer DB schema snapshot


manelcecs (approved):
This PR is applying same changes that were made in this PR: [#217](https://github.com/UN-OCHA/hpc-api/pull/217)

PR #144, opened 2024-04-10
HPC-9142: Add `parentGoverningEntityId` field to `planEntity` model

enxtur:


Pl217 (changes requested):
The work needs to be done in 2 steps, which I [wrote in the ticket](https://humanitarian.atlassian.net/browse/HPC-9142?focusedCommentId=172430). For this first step, we only need commit c8b3b862063e813a4e5afad045802cb95f29ccb2 and for the second step you can use commit acdbff0f1adad88722533b60323fd0f03698736d, but also remove `entitiesAssociation` model in the second step.

Pl217 (approved):

PR #147, opened 2024-04-19
HPC-9142: Remove `entitiesAssociation` model and its usages

enxtur:
[PR-144](https://github.com/UN-OCHA/hpc-api-core/pull/144) needs to be merged first


Pl217 (changes requested):
Instead of basing this branch off `develop`, you should have used `HPC-9142` as a base branch. The way you did it, commit c8b3b862063e813a4e5afad045802cb95f29ccb2 is duplicated. This work will need a rebase after #144 is merged. If you based this branch against `HPC-9142`, after merging the other PR, this one would auto-change its base to be `develop`.

But, regardless of that, I need to request changes here, because you didn't remove `entitiesAssociation` model, which I asked in [this review comment](https://github.com/UN-OCHA/hpc-api-core/pull/144#pullrequestreview-2009800395).

src/lib/data/planEntities.ts
```diff
@@ -257,13 +232,12 @@ const getCustomReferenceAndTypeForPlanEntity = ({
...
   );
-  const parent = entityAssociations.get(planEntity.id);
   let ref = '';
-  if (parent) {
+  if (planEntity.parentGoverningEntityId) {
     const ge = getRequiredDataByValue(
       governingEntities,
       planEntity,
-      () => parent.parentId
+      () => planEntity.parentGoverningEntityId
```
Pl217: Main purpose of method `getRequiredDataByValue` is to provide a map to lookup as first argument, entity that is used as key during lookup as second argument, and method how to fetch keys of that entity as third argument.

Read the docstring of this method for better explanation. For the third argument, docstring says:
> ... can also use a different key if entity is solely used for producing nice error messages when lookup fails.

That was the case before the changes, but we are now directly depending on a relationship between governing and plan entities, so we should write this as
```suggestion
      (pE) => pE.parentGoverningEntityId
```

In the end, this is essentially the same code functionally, but since this code can be used as model for a newly written code, it could be falsely assumed what the purpose of third argument is.

Pl217 (changes requested):

Pl217 (approved):
I think that removal of `entitiesAssociation` model should mean that we need a major version bump when we release next version of this package.

PR #152, opened 2024-07-11
HPC-8569: Add `count` function and `distinct` option to models library

manelcecs:
This PR aims to add support for common database functions, such as:

- _count_
- _distinct_
- ~~orderBy: '_nulls last_'~~ (delayed until Knex gets updated)

This changes applies to the previously defined function: '**_find_**' - and add a new function: '**_count_**'

### DIFF
--- a/src/db/util/raw-model.ts
+++ b/src/db/util/raw-model.ts
@@ -45,6 +45,16 @@ export type FindFn<F extends FieldDefinition, AdditionalArgs = {}> = (
      */
     skipValidation?: boolean;
     trx?: Knex.Transaction<any, any>;
+    /**
+     * Uses `DISTINCT ON`, which is a feature unique to PostgreSQL.
+     *
+     * Keeps only the first row of each set of rows where the given (set of) column(s) is unique.
+     * Note that the “first row” of each set is unpredictable unless `ORDER BY` is used to
+     * ensure that the desired row appears first. If ordering is used, there is an important
+     * constraint, because the first expression of `ORDER BY` must be present **somewhere** in
+     * `DISTINCT ON` expressions. But, not all expressions of `ORDER BY` need to be in `DISTINCT ON`.
+     */
+    distinct?: Array<keyof InstanceDataOf<F>>;
   } & AdditionalArgs
 ) => Promise<Array<InstanceDataOf<F>>>;
 
@@ -160,6 +170,7 @@ export const defineRawModel =
       orderBy,
       skipValidation = false,
       trx,
+      distinct,
     } = {}) => {
       const builder = trx ? masterTable().transacting(trx) : replicaTable();
       const query = builder.where(prepareCondition(where ?? {})).select('*');
@@ -177,6 +188,18 @@ export const defineRawModel =
           orderBy = [orderBy];
         }
         query.orderBy(orderBy);
+      }
+
+      if (distinct && distinct.length >= 1) {
+        if (
+          (orderBy?.length ?? 0) >= 1 &&
+          orderBy?.at(0)?.column !== distinct.at(0)
+        ) {
+          throw new Error(
+            'SELECT DISTINCT ON expressions must match initial ORDER BY expression'
+          );
+        }
+        query.distinctOn(distinct);
       }
 
       const res = await query;
--- a/src/db/util/raw-model.ts
+++ b/src/db/util/raw-model.ts
@@ -83,6 +83,20 @@ export type DestroyFn<F extends FieldDefinition> = (args: {
 
 export type TruncateFn = (trx?: Knex.Transaction<any, any>) => Promise<void>;
 
+export type CountFn<F extends FieldDefinition, AdditionalArgs = {}> = (
+  args?: {
+    where?: WhereCond<F>;
+    trx?: Knex.Transaction<any, any>;
+    /**
+     * Will produce `SELECT COUNT(DISTINCT "columnName")` and if there
+     * is an expression in `COUNT`, it will compute the number of input
+     * rows in which the input value is **not null**. So, besides selecting
+     * unique values of "columnName", it will also filter out `NULL` values
+     */
+    distinct?: Array<keyof InstanceDataOf<F>>;
+  } & AdditionalArgs
+) => Promise<bigint>;
+
 export type ModelInternals<F extends FieldDefinition> = {
   readonly type: 'single-table';
   readonly tableName: string;
@@ -102,6 +116,7 @@ export type Model<F extends FieldDefinition, AdditionalFindArgs = {}> = {
   readonly update: UpdateFn<F>;
   readonly destroy: DestroyFn<F>;
   readonly truncate: TruncateFn;
+  readonly count: CountFn<F, AdditionalFindArgs>;
 };
 
 export type InstanceDataOfModel<M extends Model<any>> = M extends Model<infer F>
@@ -252,6 +267,25 @@ export const defineRawModel =
       await builder.truncate();
     };
 
+    const count: CountFn<F> = async ({ where, trx, distinct } = {}) => {
+      const builder = trx ? masterTable().transacting(trx) : replicaTable();
+      const query = builder.where(prepareCondition(where ?? {}));
+
+      if (distinct !== undefined) {
+        query.countDistinct(distinct);
+      } else {
+        query.count('*');
+      }
+
+      const res = (await query) as unknown as [{ count: string }];
+
+      if (res.length !== 1) {
+        throw new Error('Count result must be an array of single element');
+      }
+
+      return BigInt(res[0].count);
+    };
+
     return {
       _internals: {
         type: 'single-table',
@@ -266,5 +300,6 @@ export const defineRawModel =
       update,
       destroy,
       truncate,
+      count,
     };
   };


src/db/util/raw-model.ts
```diff
@@ -45,6 +45,8 @@ export type FindFn<F extends FieldDefinition, AdditionalArgs = {}> = (
      */
     skipValidation?: boolean;
     trx?: Knex.Transaction<any, any>;
+    orderByNullsLast?: boolean;
```
Pl217: `orderByNullsLast` should be an option in `OrderByCond` type, because it is related to ordering, not directly related to `find`. What's more, if we look at [Knex docs for `orderBy` function](https://knexjs.org/guide/query-builder.html#orderby), we see that it has single and multi-column variant.

If you look at this part of our code
https://github.com/UN-OCHA/hpc-api-core/blob/8f5141594e3e75fa728f5eadc131c4314d209602/src/db/util/raw-model.ts#L175-L180
you can see that we're always using multi-column approach, because we are defining `OrderByCond` as an object:
https://github.com/UN-OCHA/hpc-api-core/blob/8f5141594e3e75fa728f5eadc131c4314d209602/src/db/util/raw-model.ts#L31-L34

Knex docs give this example:
```ts
knex('users').orderBy([
  { column: 'email' },
  { column: 'age', order: 'desc', nulls: 'last' },
]);
```
So, since we already support `column` and `order` keys, we should just add third optional key `nulls?: 'first' | 'last';` to `OrderByCond` and this feature will be fully usable with our models.

src/db/util/raw-model.ts
```diff
@@ -176,7 +189,28 @@ export const defineRawModel =
...
+          // We need to add both the columns in the order by clause and the distinct clause
+          // Merge the list of columns and keep only unique values
+          const distinctColumns = [
+            ...new Set([...distinct, ...orderBy.map((order) => order.column)]),
+          ];
+          query.distinctOn(distinctColumns);
+        } else {
+          // If there is no order by clause, we can directly use the distinct clause
+          query.distinctOn(distinct);
+        }
```
Pl217: I thought we cannot implement `DISTINCT` functionality, unless we strongly type the returned values to only have properties specified for the distinction. I didn't know about the existence of `DISTINCT ON`, which according to [Knex docs](https://knexjs.org/guide/query-builder.html#distincton) is a unique feature of PostgreSQL.

On [PostgreSQL docs page](https://www.postgresql.org/docs/current/sql-select.html), under the "DISTINCT Clause" section (cannot link directly to that section), we can read how this feature operates:

> `SELECT DISTINCT ON ( expression [, ...] )` keeps only the first row of each set of rows where the given expressions evaluate to equal. The `DISTINCT ON` expressions are interpreted using the same rules as for `ORDER BY` (see above). Note that the “first row” of each set is unpredictable unless `ORDER BY` is used to ensure that the desired row appears first. For example:
> ```sql
> SELECT DISTINCT ON (location) location, time, report
>    FROM weather_reports
>    ORDER BY location, time DESC;
> ```
> retrieves the most recent weather report for each location. But if we had not used `ORDER BY` to force descending order of time values for each location, we'd have gotten a report from an unpredictable time for each location.
>
> The `DISTINCT ON` expression(s) must match the leftmost `ORDER BY` expression(s). The `ORDER BY` clause will normally contain additional expression(s) that determine the desired precedence of rows within each `DISTINCT ON` group.

Again looking at [Knex docs](https://knexjs.org/guide/query-builder.html#distincton), we see that it takes form `.distinctOn([*columns])`, so I tried writing this:
```ts
const p = await models.v4.participant.find({
  distinct: ['name'],
});
```
which produces this SQL:
```sql
SELECT DISTINCT ON ("name") *
FROM "participant"
```
So, the column names we use on `distinct` will eventually be passed to `distinctOn` method from Knex, which just adds these columns to the parenthesis and you cannot control what comes after parenthesis, it's always `*`, which is actually what I would want, since we don't need to strongly type which columns are returned, since all are returned always, just like with regular `SELECT`, without `DISTINCT`.

I played around a lot with `SELECT DISTINCT ON` and better understood what they mean in PostgreSQL docs when they say "The `DISTINCT ON` expression(s) must match the leftmost `ORDER BY` expression(s)". It isn't necessary to use `ORDER BY` with `DISCTINCT ON`, but if you care which row is returned when multiple rows have the same column (if you only use a single column in `DISTINCT ON`), then you also need ordering.

So, I wanted to get all distinct names of users in our system and by grouping, I found that there are 9 instances where name is "n/a n/a". This is strange "name", but let's not focus on that. What I wanted to then achieve by using `DISTINCT ON` is to get the user with the lowest ID. So, I attempted writing this SQL:
```sql
SELECT DISTINCT ON (p."name") *
FROM participant p
ORDER BY p.id;
```
But, when I tried executing it, I've got error `SELECT DISTINCT ON expressions must match initial ORDER BY expressions`. To achieve what I wanted, I needed to:
1) Include `p.id` in `DISTINCT ON`. But, this will change what we are considering distinct rows and no longer filter out duplicate names. Since I am including ID (which is unique), this won't filter anything
2) Include `p."name"` in ordering, but it needs to be first expression. I don't need sorting by name, but due to constraints, first expression of `ORDER BY`, must be present **somewhere** in `DISTINCT ON` expressions

So, with the modified SQL, I achieved what I wanted:
```sql
SELECT DISTINCT ON (p."name") *
FROM participant p
ORDER BY p."name", p.id;
```
The important thing to notice here is that **not all expressions of `ORDER BY` need to be in `DISTINCT ON`**. Especially if you strictly append them at the end of `DISTINCT ON` list, you are changing how distinction is made and preventing the library user from writing queries they want.

This feature `DISTINCT ON` needs developers to know what they are doing, and they must know the implications of queries they write. I don't know how often it would be important to get the specific row "representing" a distinct value. If not often, library clients will write the code without ordering. But, when ordering is needed, we need to respect the PostgreSQL constraint. We can leave it to PostgreSQL to fail or we can fail in library code ourselves. I would add a check in our library code, just so that the limitation becomes more obvious:
```suggestion
      if (distinct && distinct.length >= 1) {
        if (
          (orderBy?.length ?? 0) >= 1 &&
          orderBy?.at(0)?.column !== distinct.at(0)
        ) {
          throw new Error(
            'SELECT DISTINCT ON expressions must match initial ORDER BY expression'
          );
        }

        query.distinctOn(distinct);
      }
```

src/db/util/raw-model.ts
```diff
@@ -45,6 +45,8 @@ export type FindFn<F extends FieldDefinition, AdditionalArgs = {}> = (
      */
     skipValidation?: boolean;
     trx?: Knex.Transaction<any, any>;
+    orderByNullsLast?: boolean;
+    distinct?: Array<keyof InstanceDataOf<F>>;
```
Pl217, line 57: I'd document that this field uses `DISTINCT ON` clause, since it is specific to PostgreSQL and different from plain `DISTINCT`, which exists in SQL standard.
```suggestion
    /**
     * Uses `DISTINCT ON`, which is a feature unique to PostgreSQL.
     *
     * Keeps only the first row of each set of rows where the given (set of) column(s) is unique.
     * Note that the “first row” of each set is unpredictable unless `ORDER BY` is used to
     * ensure that the desired row appears first.
     */
    distinct?: Array<keyof InstanceDataOf<F>>;
```

src/db/util/raw-model.ts
```diff
@@ -229,6 +263,25 @@ export const defineRawModel =
...
 
+    const count: CountFn<F> = async ({ where, trx, distinct } = {}) => {
+      const builder = trx ? masterTable().transacting(trx) : replicaTable();
+      const query = builder.where(prepareCondition(where ?? {}));
+
+      if (distinct !== undefined) {
+        query.countDistinct(distinct);
+      } else {
+        query.count('*');
+      }
```
Pl217, line 278: I was looking at [PostgreSQL docs](https://www.postgresql.org/docs/14/functions-aggregate.html) and played around to see how different queries behave. I tested `COUNT(*)` and `COUNT("columnName")` to see how it behaves and whether we need to support only `*` or column names too, since [Knex docs](https://knexjs.org/guide/query-builder.html#count) always seem to use column names.

Turns out, when you use `count(expression)`, it [will return](https://www.postgresql.org/docs/8.2/functions-aggregate.html) (old docs are more intuitive to me):
> number of input rows for which the value of `expression` is not null
So, the difference between
```sql
SELECT COUNT("name")
FROM "participant";
```
and
```sql
SELECT COUNT(DISTINCT "name")
FROM "participant";
```
is that first one will return number of rows where `name IS NOT NULL`. The second one will return number of **distinct** names which, again, **are not null**.

Since, we can replicate the behavior of `COUNT("name")` by using `COUNT(*)` and `WHERE "name" IS NOT NULL`, we are fine with only allowing `*`.

We can also keep the option to use `distinct`, but we should document that this expression will eliminate `NULL` values from count, because that is important quirk and could become [gotcha](https://en.wikipedia.org/wiki/Gotcha_(programming)).

src/db/util/raw-model.ts
```diff
@@ -73,6 +75,14 @@ export type DestroyFn<F extends FieldDefinition> = (args: {
 
 export type TruncateFn = (trx?: Knex.Transaction<any, any>) => Promise<void>;
 
+export type CountFn<F extends FieldDefinition, AdditionalArgs = {}> = (
+  args?: {
+    where?: WhereCond<F>;
+    trx?: Knex.Transaction<any, any>;
+    distinct?: Array<keyof InstanceDataOf<F>>;
```
Pl217, line 96: Like I said in the other comment, I think we should document that `COUNT` filters out `NOT NULL` values if expression is provided instead of `*`
```suggestion
    /**
     * Will produce `SELECT COUNT(DISTINCT "columnName")` and if there
     * is an expression in `COUNT`, it will compute the number of input
     * rows in which the input value is **not null**. So, besides selecting
     * unique values of "columnName", it will also filter out `NULL` values
     */
    distinct?: Array<keyof InstanceDataOf<F>>;
```

src/db/util/raw-model.ts
```diff
@@ -73,6 +75,14 @@ export type DestroyFn<F extends FieldDefinition> = (args: {
...
+  } & AdditionalArgs
+) => Promise<number>;
```
Pl217: Since PostgreSQL returns `bigint` type, type which will be represented as a string and not a number, as can be seen in both [PostgreSQL](https://www.postgresql.org/docs/16/functions-aggregate.html) and [Knex docs](https://knexjs.org/guide/query-builder.html#count), I would convert back the string to [`BigInt` type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), since it now natively exists in Javascript.
```suggestion
) => Promise<bigint>;
```

src/db/util/raw-model.ts
```diff
@@ -229,6 +263,25 @@ export const defineRawModel =
...
+        query.count('*');
+      }
+
+      const res = await query;
+
+      const mappedResponse = res as Array<{ count: string }>;
+
+      return mappedResponse.length > 0
+        ? parseInt(mappedResponse[0].count, 10)
+        : 0;
```
Pl217: Since PostgreSQL returns `bigint` type, type which will be represented as a string, I would type the response even stronger, as a tuple (array with single element) that contains count object.

Next, I would not try to convert it back to a number, because there is a valid reason it is a string. So, let's convert to a `BigInt` and don't return 0 as a default value, especially if response is invalid (has zero elements in your case).

Let's rewrite like this
```ts
const res = (await query) as unknown as [{ count: string }];

if (res.length !== 1) {
  throw new Error('Count result must be an array of single element');
}

return BigInt(res[0].count);
```
or use `io-ts` to validate the response fully, not only that it has a single element
```ts
const COUNT = t.tuple([
  t.type({
    count: t.string,
  }),
]);

const decode = COUNT.decode(res);

if (!isRight(decode)) {
  const errors = ioTsErrorFormatter(decode);
  throw new DataValidationError({
    errors,
    identifier: 'count query',
    data: res,
  });
}

return BigInt(decode.right[0].count);
```
I think we should use the first, simpler option, because response doesn't really have much room to violate the shape.

Pl217 (changes requested):
Thanks for adding these new capabilities to our model library. New options are awesome!

I didn't even know about some possibilities, like `DISTINCT ON`, which made me read a lot of documentation and play around with PostgreSQL. I did spend entire afternoon researching about various options and thinking about our needs, much more than I anticipated when I started review :/

My responses are long and detailed, but addressing my review comments should take little time.

src/db/util/raw-model.ts
manelcecs: I did not found that example because that is what we needed - definitely I'll be adding that change.

src/db/util/raw-model.ts
manelcecs: I think that the first option is the best - There is no need to add that level of validation for a simple query like this one.

src/db/util/raw-model.ts
Pl217: Crap, we cannot use `nulls` option, because it only exists in newer versions of Knex. The feature was added by [this PR](https://github.com/knex/knex/pull/4720), which was [released in `v0.95.12` of Knex](https://github.com/knex/knex/releases/tag/0.95.12).

There was also [a bug](https://github.com/knex/knex/issues/4837) with the implementation in PostgreSQL, which was resolved in [this PR](https://github.com/knex/knex/pull/4989). That was [released in `v1.0.2`](https://github.com/knex/knex/releases/tag/1.0.3), so we may need to update to at least that version.

I think we should update now and [HPC-8383](https://humanitarian.atlassian.net/browse/HPC-8383) should be a blocker for this work, because we are long overdue to update Knex, since we are at `v0.21.1`, released 28 April, 2020!

src/db/util/raw-model.ts
```diff
@@ -45,6 +46,14 @@ export type FindFn<F extends FieldDefinition, AdditionalArgs = {}> = (
      */
     skipValidation?: boolean;
     trx?: Knex.Transaction<any, any>;
+    /**
+     * Uses `DISTINCT ON`, which is a feature unique to PostgreSQL.
+     *
+     * Keeps only the first row of each set of rows where the given (set of) column(s) is unique.
+     * Note that the “first row” of each set is unpredictable unless `ORDER BY` is used to
+     * ensure that the desired row appears first.
+     */
```
Pl217: I know it was my suggestion to add this docstring, but now I realize there is an important constraint related to ordering that we should document
```suggestion
    /**
     * Uses `DISTINCT ON`, which is a feature unique to PostgreSQL.
     *
     * Keeps only the first row of each set of rows where the given (set of) column(s) is unique.
     * Note that the “first row” of each set is unpredictable unless `ORDER BY` is used to
     * ensure that the desired row appears first. If ordering is used, there is an important
     * constraint, because the first expression of `ORDER BY` must be present **somewhere** in
     * `DISTINCT ON` expressions. But, not all expressions of `ORDER BY` need to be in `DISTINCT ON`.
     */
```

src/db/util/raw-model.ts
```diff
@@ -229,6 +266,25 @@ export const defineRawModel =
...
+      const builder = trx ? masterTable().transacting(trx) : replicaTable();
+      const query = builder.where(prepareCondition(where ?? {}));
+
+      if (distinct !== undefined) {
+        query.countDistinct(distinct);
+      } else {
+        query.count('*');
+      }
+
+      const res = (await query) as unknown as Array<{ count: string }>;
```
Pl217: In my [suggested change](https://github.com/UN-OCHA/hpc-api-core/pull/152#discussion_r1674313558) from previous round of reviews, I have suggested using
```ts
const res = (await query) as unknown as [{ count: string }];
```
but you defined it using `Array`. I want to point out the difference. I have suggested using a **tuple**, which specifies this as an **array of single element**, which has object with `count` property. When you use `Array`, you indicate there can be multiple objects in an array, each having `count` property. You can see the distinction if you hover `res.length` on line 281, TS says that its type is `number`. If you change the code to use a tuple, as I suggested, the type will be `1`.

In this case, typing it as a tuple is more realistic, because that is what Knex returns, but it doesn't bring any difference to the code, as this is purely typing thing. Just wanted to explain the difference, you may not want to use a tuple for some reason.

Pl217 (changes requested):
We cannot add `nulls` option until we update to a more recent version of Knex, as I wrote [in this comment](https://github.com/UN-OCHA/hpc-api-core/pull/152#discussion_r1679359219).

If you want you can drop the first commit which adds `nulls` option and merge the rest (after addressing really minor comments I have inline now), but you may also want to block the entire work on HPC-8569 until Knex is updated, because you cannot add `nulls last` to your queries.

Discuss with Kashif (once he's back next week) whether we want to update Knex now and block this work. Otherwise, I would implement `nulls last` functionality in code and write a TODO comment to refactor it once we can add that option in models library.

src/db/util/raw-model.ts
manelcecs: Done!

manelcecs:
> We cannot add `nulls` option until we update to a more recent version of Knex, as I wrote [in this comment](https://github.com/UN-OCHA/hpc-api-core/pull/152#discussion_r1679359219).
> 
> If you want you can drop the first commit which adds `nulls` option and merge the rest (after addressing really minor comments I have inline now), but you may also want to block the entire work on HPC-8569 until Knex is updated, because you cannot add `nulls last` to your queries.
> 
> Discuss with Kashif (once he's back next week) whether we want to update Knex now and block this work. Otherwise, I would implement `nulls last` functionality in code and write a TODO comment to refactor it once we can add that option in models library.

@Pl217 After discuss this issue with Kashif - We have decided to drop the commit and after the implementation of the new version of Knex we update both HPC-API-CORE in order to add the 'nulls last' functionality and after, update HPC-API to add the order to the needed queries.

Pl217 (approved):

PR #154, opened 2024-08-23
HPC-8383: Upgrade knex to latest version

Pl217:
The goal of this PR is to update Knex to latest version. The update is done in stages, for two reasons:
1. Significant adjustments need to be done on the codebase and they can be reviewed separately. So, even if it isn't a major release, nor it would seem relevant if you look at [migration docs](https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/UPGRADING.md), version `v0.21.12` required most of my time. In the rest of my comment, I will mostly focus on decisions made in this commit. The end result looks like cheating (as I used a bunch of `as any`), but it was maybe the only solution. Subsequent updates are to latest version of each major version. In `v0.95.15` there are most breaking changes, but it seems to me that only import style changes affect us. In `v1.0.7` there is last significant type definition change that affected us
2. Some upgrades, even major version upgrades could be squashed and we could realistically go from `v1.0.7` to latest version `v3.1.0`, but I still decided to split them to make potential troubleshooting using `git bisect` easier. All commits are passing TS type checks

Now I will focus on what I spent most of my time figuring out. In Knex `v0.21.12`, the type definitions changed significantly, which affected our custom model library

If you were to diff [Knex `v0.21.11` type definitions](https://github.com/knex/knex/blob/4cfb3ea17b8580d370f096807b7f8954a32a5f89/types/index.d.ts) and [Knex `v0.21.12` type definitions](https://github.com/knex/knex/blob/90b145dc84b1281b985fa5cedd09b8cb362add9e/types/index.d.ts), you will see that they have introduced the following generic types:
```ts
export type CompositeTableType<
  TBase,
  TInsert = TBase,
  TUpdate = Partial<TInsert>
> = {
  base: TBase;
  insert: TInsert;
  update: TUpdate;
};

type TableInterfaceScope = keyof CompositeTableType<unknown>;

type ResolveTableType<
  TCompositeTableType,
  TScope extends TableInterfaceScope = 'base'
> = TCompositeTableType extends CompositeTableType<unknown>
  ? TCompositeTableType[TScope]
  : TCompositeTableType;
```
The changes were introduced in [this PR](https://github.com/knex/knex/pull/4071), which was released in `v0.21.10` and reverted in `v0.21.11`, then reintroduced in [this PR](https://github.com/knex/knex/pull/4100), and re-released in `v0.21.12`.

The goal of the feature was to [allow to globally define types for table columns](https://knexjs.org/guide/#typescript) and the feature is opt-in, as it falls back to old behavior if `Tables` interfaces is not extended and it supports all old use-cases.

But, this new generic introduced a problem with our custom library, which was written around Knex.

Namely, the definitions of methods started using `CompositeTableType` and `ResolveTableType`, so for example, `insert` function changed from:

```ts
insert<TResult2 = number[]>(
  data: DbRecordArr<TRecord>
): QueryBuilder<TRecord, TResult2>;
```
to
```ts
insert<TResult2 = number[]>(
  data: TRecord extends CompositeTableType<unknown>
    ? ResolveTableType<TRecord, 'insert'>
    : DbRecordArr<TRecord>
): QueryBuilder<TRecord, TResult2>;
```

We can see the conditional type is now used, trying to match the shape of `CompositeTableType` and use old behavior if not matching. Since we're not using global table definitions, but defining our own library, the problem occurs where Typescript cannot infer that our custom types are not matching the shape of `CompositeTableType<unknown>`.

The problem is that Knex is using a conditional type now, which can be illustrated if we change it to:
```ts
insert<TResult2 = number[]>(
  data: TRecord extends CompositeTableType<unknown>
    ? DbRecordArr<TRecord>
    : DbRecordArr<TRecord>
): QueryBuilder<TRecord, TResult2>;
```

i.e. both "branches" of conditional type are same, but the error is still there.

Since our custom types don't match the shape of `CompositeTableType`, the conditional type `ResolveTableType` should, in theory, resolve to the "else" branch (i.e., `TCompositeTableType`).

But, the issue here is related to TypeScript's limitations when it comes to resolving conditional types with complex generic parameters. In our case, the problem arises because the type parameter `TCompositeTableType` is not known at the point where TypeScript needs to resolve the conditional type.

Let's create a simplified example that mimics the situation:
```ts
// Simplified version of Knex's types
interface CompositeTableType<T> {
  base: T;
  insert: Partial<T>;
  update: Partial<T>;
}

type ResolveTableType<
  TCompositeTableType,
  TScope extends 'base' | 'insert' | 'update' = 'base'
> = TCompositeTableType extends CompositeTableType<unknown>
  ? TCompositeTableType[TScope]
  : TCompositeTableType;

// Simplified version of our model custom types
interface FieldDefinition {
  required: Record<string, unknown>;
  optional: Record<string, unknown>;
}

type InstanceDataOf<F extends FieldDefinition> = F['required'] &
  Partial<F['optional']>;

type UserDataOf<F extends FieldDefinition> = F['required'] &
  Partial<F['optional']>;

function createModel<F extends FieldDefinition>() {
  return {
    create: async (data: UserDataOf<F>): Promise<InstanceDataOf<F>> => {
      // Simulating Knex's insert operation
      // We get an error: "Type `UserDataOf<F>` is not assignable to type
      // `ResolveTableType<InstanceDataOf<F>, "insert">`"
      const result: ResolveTableType<InstanceDataOf<F>, 'insert'> = data;
      return result;
    },
  };
}

// In concrete usage, Typescript can infer the type
const model = createModel<{
  required: { id: number; name: string };
  optional: { email: string };
}>();

const result = model.create({ id: 1, name: 'John' });
```

In this example, TypeScript cannot definitively resolve the `ResolveTableType` conditional type because:
1. The `InstanceDataOf<F>` type is not fully known at the point where `ResolveTableType` is used.
2. TypeScript can't guarantee that `InstanceDataOf<F>` will never satisfy `CompositeTableType<unknown>`, even though we know it won't.

> [!IMPORTANT]  
> This limitation occurs because TypeScript evaluates conditional types with type parameters in a deferred manner. It doesn't immediately substitute the actual types, which would allow it to definitively resolve the condition.

To work around this limitation, we need to provide explicit type annotations or use type assertions at certain points in our code, effectively telling TypeScript which branch of the conditional type to use. But, we need to resort to using `as any` type assertions, which are far from ideal, and should really be avoided, but there is no option for more type-safe type assertion here, because **Knex doesn't export the needed types**.

Usage of `as any` assertion isn't a huge problem in this case, because we're not really relying on Knex type definitions, but we have a wrapper around Knex, and all the strong typing happens in this wrapper library. The strong typing when calling methods from Knex was a nice addition in overall type safety, but we're not losing much, besides the clarity which method override is being matched exactly, but that was little convoluted already, as Ctrl+Click (or F12 - Go to Definition) wasn't really leading to the exact override in the previous version either.

Nevertheless, `as any` assertion was used here as a last resort, because there are no other viable alternatives. I was also thinking about overriding Knex's types and brining back old behavior, but it isn't so straightforward to override identifiers by using the same name, because we'd need a new (dev) dependency `patch-package` to help us, and then we would have maintenance overhead to keep this override correct in each version of Knex.


enxtur (approved):
What a nice practice that upgrade versions in stages and resolve issues in each upgrade. I wouldn't think that way. I tried, but couldn't figure out issue you described.

Onitoxan:
@Pl217 I have been trying different ways to make it work and tested then, I could just think of these 3 solutions, but they come with problems:

1. Casting the actual type, In the little example you provided is easy to do, but checking out in the codebase the actual type, it may be even worse because of how complicated it could be to write and confusing for someone working on this code

2. Typeguards, the only thing with this approach is that it will be adding some new code and nesting just for TS to properly infer the type, so Im not sure this is beneficial at all. Plus, writing this Typeguards will not be easy

3. As you said in your comment, overriding the type definitions, but I'm not sure is worth the maintenance cost for the value we will get out of it

Since we have our own library built on top of the old version of Knex I think the remedy of adding string typing will be worse than the problem itself. So in my opinion, casting `as any` is actually the best solution in this case, being used as last resort.

Pl217:
@Onitoxan thanks for the suggestions, but as I said, the problem is that Knex isn't exporting the types I need to do any safer type assertions and cannot easily be recreated either, due to lots of interdependency. All of the solutions I can think of would involve maintenance cost, which I want to aviod.

PR #156, opened 2024-08-29
Update dependencies & enable Dependabot version updates

Pl217:


enxtur (changes requested):
It seems that only `src/db/models/authTarget.ts` has a linting issue, even though it hasn’t been modified. Should we consider applying linting checks to only to the touched files besides fix this issue?

Pl217:
> It seems that only `src/db/models/authTarget.ts` has a linting issue, even though it hasn’t been modified. Should we consider applying linting checks to only to the touched files besides fix this issue?

I have fixed the linting error and updated packages which got patch version updates since opening of this PR (`@types/node` and `lint-staged`) to avoid Dependabot opening PRs to update them immediately after merging this PR, because version updates run monthly as configured, but also after each change to `dependabot.yml` file.

Regarding what we want to lint, checking linting and formatting in entire repository is exactly what we want actually. Because doing so caught a problem that wouldn't be caught if we check only changed files. In other repos, we check only changed files because there are existing problems in old legacy files which we didn't spend time resolve.

enxtur (approved):

PR #176, opened 2024-10-29
HPC-9832: Make `destroy` supports `softDeletionEnabled`

enxtur:
From [this](https://github.com/UN-OCHA/hpc_service/pull/3571#pullrequestreview-2395741396) comment, I was initially trying to add `hardDelete` option to destroy method in `src/db/util/sequelize-model.ts`. But appears that it wasn't even support `softDeletionEnabled` option.


enxtur:
Marking it as unnecessary

PR #187, opened 2024-11-14
HPC-8569 get table columns function

manelcecs:


Delgee (changes requested):
Since `getTableColumns` is runtime version of https://github.com/UN-OCHA/hpc-api-core/blob/dc5888a7a3617adc95c41cafbc3d3da863c0b5fa/src/db/util/model-definition.ts#L74
can we do it similar fashion instead of looping through every child.

manelcecs:
> Since `getTableColumns` is runtime version of
> 
> https://github.com/UN-OCHA/hpc-api-core/blob/dc5888a7a3617adc95c41cafbc3d3da863c0b5fa/src/db/util/model-definition.ts#L74
> 
> 
> can we do it similar fashion instead of looping through every child.

Would this code be a better approach?
```
export const getTableColumns = <
  T extends { _internals: { fields: FieldDefinition } },
>(
  table: T
): string[] => {
  const {
    generated,
    generatedCompositeKey,
    nonNullWithDefault,
    required,
    optional,
    accidentallyOptional,
  } = table._internals.fields;

  // Collect columns from all subsets
  const columns = [
    ...Object.keys(generated ?? {}),
    ...Object.keys(generatedCompositeKey ?? {}),
    ...Object.keys(nonNullWithDefault ?? {}),
    ...Object.keys(required ?? {}),
    ...Object.keys(optional ?? {}),
    ...Object.keys(accidentallyOptional ?? {}),
  ];

  return columns;
};
```

Delgee:
I believe so.

Delgee (approved):

PR #198, opened 2024-12-05
HPC-9162: OneGMS bridge

Delgee:


manelcecs (approved):
Not a big change tho!

PR #201, opened 2024-12-25
Clean up entity prototype codec

Pl217:
After cleaning the invalid data in DB with command from https://github.com/UN-OCHA/hpc_service/pull/3683, the codec no longer needs to accommodate string IDs and key with typo "cadinality", and therefore `cardinality` can be made required


enxtur (approved):

Pl217:
The admin command from https://github.com/UN-OCHA/hpc_service/pull/3683 has been merged now, but it needs to be deployed + run on production, and then we can merge this PR.

PR #210, opened 2025-01-06
HPC-9824: Coordinating organization information gets deleted from plan details

enxtur:


manelcecs (approved):
I believe everything is fine but also I think there is a OCD with sorting this kind of 'hardcoded' labels and it ment to be alphabetically sorted.
I don't really mind tbf - but just checking.

Also, there is a conflict with the `package.json`, please fix it.

PR #212, opened 2025-01-20
HPC-9944: Add v4 models to `Role` and `WorkflowRole` tables

Delgee:


src/db/models/workflowRole.ts
```diff
@@ -2,11 +2,39 @@ import * as t from 'io-ts';
...
+import { ROLE_ID } from './role';
 export type WorkflowRoleId = Brand<
   number,
   { readonly s: unique symbol },
   'workflowRole.id'
 >;
 
 export const WORKFLOW_ROLE_ID = brandedType<number, WorkflowRoleId>(t.number);
+
+// export const PERMITTED_ACTION_IDS = t.union([
```
manelcecs, line 15: Remove this unused code

manelcecs (changes requested):
Please review the comment! Other than that is looking fine

Pl217 (changes requested):
The reason why v4 models don't have these tables is that they are considered legacy and should not be used in any new code. Since you need to write **temporary** admin command, which must fix corrupt legacy data, I think you can use legacy models for that purpose.

In [related hpc_service PR](https://github.com/UN-OCHA/hpc_service/pull/3712), if you can write all code using v4 models and only use legacy models for legacy tables, that would be best. But, if combining models isn't possible in some place, it's fine to use legacy models there too, since it is a temporary thing, fixing legacy stuff.

Delgee:
No longer needed.

PR #218, opened 2025-02-05
HPC-10029: Optimize `toCamelCase`

Pl217:
In the ticket, it was reported that call to HPC API endpoint `v2/plan/1185/responseMonitoring?includeCaseloadDisaggregation=true&includeIndicatorDisaggregation=false&disaggregationOnlyTotal=false` doesn't complete until timeout is reached. After investigation using profiling tool of DevTools, I found that calls to `toCamelCase` are actually responsible to great extent.

First, I have tried various ways of optimizing the method itself, to only have managed getting from ~45 seconds to ~25 seconds while processing attachment data on HPC API side. Without calls to `toCamelCase`, that code took 5 seconds. Even Lodash's implementation was worse than this, at around 35 seconds, all measured on local. So, just optimization didn't cut it, so I decided to "cache" the responses of this method, which brought down number of its invocations from 16,850,784 to just 4,584! This huge reduction helped tremendously with performance.

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -278,21 +278,13 @@ export const cleanNumberVal = (value: number | string | null): number | null =>
       ? parseFloat(value.trim().replaceAll(',', ''))
       : null;
 
-export const toCamelCase = (originalString: string) => {
-  let isNextLetterUppercase = false;
-  let convertedString = '';
-
-  for (const char of originalString) {
-    if (char === ' ') {
-      isNextLetterUppercase = true;
-      continue;
-    }
-
-    convertedString += isNextLetterUppercase
-      ? char.toUpperCase()
-      : char.toLowerCase();
-    isNextLetterUppercase = false;
-  }
-
-  return convertedString;
-};
+export const toCamelCase = (originalString: string) =>
+  originalString
+    .trim()
+    .split(' ')
+    .map((word, index) =>
+      index === 0
+        ? word.toLowerCase()
+        : word.charAt(0).toUpperCase() + word.slice(1).toLowerCase()
+    )
+    .join('');
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -84,6 +84,7 @@ if [ $KEEP -eq 0 ]; then
 fi
 
 yarn jest "$COMMAND_ARGS" $FORCE_STOP_JEST
+TEST_EXIT_CODE=$?
 
 if [ $KEEP -eq 0 ]; then
   # Stop Docker containers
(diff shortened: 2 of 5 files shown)


src/util/index.ts
```diff
@@ -278,21 +278,12 @@ export const cleanNumberVal = (value: number | string | null): number | null =>
...
-};
+export const toCamelCase = (originalString: string) =>
+  originalString
+    .split(' ')
+    .map((word, index) =>
+      index === 0
+        ? word.toLowerCase()
+        : word.charAt(0).toUpperCase() + word.slice(1).toLowerCase()
+    )
+    .join('');
```
enxtur: Without `.trim` its test case failing.
 
Replacing multiple spaces by single space could save some.
```js
'multiple      spaces'.split(' ') // [ 'multiple', '', '', '', '', '', 'spaces' ]
```

```suggestion
export const toCamelCase = (originalString: string) =>
  originalString
    .trim()
    .replace(/\s+/g, ' ')
    .split(' ')
    .map((word, index) =>
      index === 0
        ? word.toLowerCase()
        : word.charAt(0).toUpperCase() + word.slice(1).toLowerCase()
    )
    .join('');
```

enxtur (changes requested):
`Unit Tests` job is not failing when `./bin/test.sh` fails

Suggestion for `ci.yml`
```yaml
      - name: Run Unit Tests
        run: |
          set -e
          ./bin/test.sh
```

src/util/index.ts
Pl217: I was unable to run unit tests on my machine when I was opening the PR, since all test cases were failing, so when I opened a PR and saw that it passed, I though it's all good. Thanks for catching the issue. I will answer to both inline and code review comment here.

For failing unit tests, I'd do the same thing we're doing in hpc_service, where we capture `jest` exit code and then exit the script with that code. I think it is a responsibility of the script to do this correctly, rather than CI to account for shell script's misbehavior.

As for the `toCamelCase` method, we indeed need `trim()`, because leading spaces would be treated as first character with the current algorithm. On the other hand, I don't know whether replacing multiple spaces by single space saves time, because the call to `replace` is not free, although I don't know what is its time and space complexity. But, if it is O(n), maybe we don't save much.

I have made optimizations to `toCamelCase` method by measuring execution times, which I have done again. I needed to revert the "caching" which actually brings the significant boost in performance. Turns out that using `replace` adds 7-9 seconds to execution time. I have measured without any changes to my proposed `toCamelCase` method, then with `trim()` added and with `replace` added after that. Each of these were measured three times. Trimming doesn't add anything significantly, but `replace` impacts performance.

Now, as I said, this will not be felt because of "caching", but since method `toCamelCase` is being optimized in one of the commits, I will not add `replace`, since it hurts performance.

enxtur (approved):
I was only concerned about the unnecessary loop and didn’t realize that the regex replacing had an even greater impact :P

PR #222, opened 2025-02-20
HPC-9636: Add soft deletion support to destroy

enxtur:


Pl217 (changes requested):
I am surprised this isn't implemented before. You should define a `destroy` wrapper in `defineSequelizeModel` instead of changing the one in raw model. The `CONTRIBUTING.md` has a good documentation and I suggest reading all of it, but this part is interesting here:
https://github.com/UN-OCHA/hpc-api-core/blob/34c5bcc489b76f55d53f02e9afadcdf12b6e3912/CONTRIBUTING.md?plain=1#L151-L154

enxtur:
> I am surprised this isn't implemented before. You should define a `destroy` wrapper in `defineSequelizeModel` instead of changing the one in raw model.

@Pl217  I tried to implement as you said, but couldn't figure out other than force type casting like this.
https://github.com/UN-OCHA/hpc-api-core/blob/b0efad3796521d7763141ae5406ddebe3f9f956c/src/db/util/sequelize-model.ts#L191-L193

src/db/util/sequelize-model.ts
```diff
@@ -182,12 +183,27 @@ export const defineSequelizeModel =
...
+            ...args,
+            values: {
+              deletedAt: masterConn.fn.now(3),
+            } as unknown as Partial<UserDataOf<Fields>>,
+          })
+        ).length;
+      }
+      return await model.destroy(args);
+    };
+
```
manelcecs: I think this could be a good solution for this.
Using 'function overload' while defining a new destroy function type could help on having this approach, validate the 'values' and the needs from the update function.
```suggestion
    const destroy: DestroySequelizeFn<Fields> = async (args) => {
      if (
        opts.softDeletionEnabled &&
        'values' in args
      ) {
        return (
          await model.update({
            ...args,
            values: {
              ...args.values,
              deletedAt: masterConn.fn.now(3),
              updatedAt: masterConn.fn.now(3),
            },
          })
        ).length;
      }
      return await model.destroy(args);
    };
```

src/db/util/sequelize-model.ts
```diff
@@ -182,12 +183,27 @@ export const defineSequelizeModel =
       });
     };
```
manelcecs, line 199: We will need to add a new type function definition in order to do this approach.
```suggestion
export type DestroySequelizeFn<F extends FieldDefinition> = {
  (args: { where: WhereCond<F>; trx?: Knex.Transaction<any, any> }): Promise<number>;
  (args: { values: Partial<UserDataOf<F>>; where: WhereCond<F>; trx?: Knex.Transaction<any, any> }): Promise<number>;
};
```

manelcecs (changes requested):
Give it a think!
I believe making changes into models is more than justify when we do this kind of implementations (like this that was never implemented before)
But also having type casting is not good so making this approach will satisfy both.

src/db/util/sequelize-model.ts
manelcecs: It may not be needed to`export` this type function. It's an 'extra' from copy-pasting

src/db/util/sequelize-model.ts
enxtur: Under what circumstances would the destroy function ever be invoked with `values` within the args parameter? :P
I got an idea while looking at your suggestion. It could be a little better. Thanks for the hint and efforts.

manelcecs: It's because you where using the 'update' function from the Sequelize wrapper, so I keep it. 
Glad it help

src/db/util/sequelize-model.ts
```diff
@@ -182,12 +183,38 @@ export const defineSequelizeModel =
...
 
+    const isSoftDeletionValues = <F extends FieldDefinition>(
```
manelcecs, line 200: Having this function defined here would not make accesible when using any models?

manelcecs (approved):
All good!

src/db/util/sequelize-model.ts
```diff
@@ -182,12 +183,38 @@ export const defineSequelizeModel =
...
+
+    const destroy: DestroyFn<Fields> = async (args) => {
+      const values = {
+        deletedAt: masterConn.fn.now(3),
+      };
+      if (opts.softDeletionEnabled && isSoftDeletionValues<Fields>(values)) {
+        const res = await model.update({
+          ...args,
+          values,
+        });
```
Pl217: I'd like to point out the difference between call to `update` (defined above and `model.update` (defined in `raw-model.ts`). The former will also update the date in `updatedAt` column at the time of doing soft deletion, whereas the latter will not do that.

It's a matter of preference if we want to also update `updatedAt` value when setting `deletedAt` value, but I would personally not do it. You make the decision. Not sure if you considered both options or didn't think about it.

Pl217 (changes requested):
As I said in my last comment, I was surprised this wasn't implemented before, but there was actually already an attempt you made in #176 around the `destroy` wrapper. But, it was never reviewed and later closed since it was not needed. If we trace how that previous attempt came to be, we can find this [review comment](https://github.com/UN-OCHA/hpc_service/pull/3571#pullrequestreview-2395741396).

In there, I say that we need to soft delete some records and hard delete others. Then, I say that in order to hard delete tables which are usually soft deleted, you need to add an option to the models. Turns out, this is completely wrong, as our v4 models have been doing the hard deletion in all tables the entire time!

This is a big implication and after we're done making changes to the models, we should go back and revisit all the places we're using `destroy` from new models, because there was false assumption that it handles soft deletions for us. Delgee also pointed out in [his comment](https://github.com/UN-OCHA/hpc_service/pull/3571#issuecomment-2449483336) on the PR I previously linked that models don't even do soft deletion, but I seem to not have been involved in that PR by the time he left the comment and never properly thought about the implications of hard deleting everything.

In [that work](https://github.com/UN-OCHA/hpc_service/blob/f659f08d9a49b8829db01b59b08d0e32025edb3d/app/commands/commands/delete-broken-attachment.ts#L144-L180), you have accounted for difference between soft deletion and hard deletion. I was worried to set `updatedAt` with `new Date()` because I thought it may have higher precision than we usually have, since we [needed to strip](https://github.com/UN-OCHA/hpc-api-core/blob/7bf3c460c7d3b2a1adaab411639470d382ba3171/src/db/util/sequelize-model.ts#L180) knex to 3 millisecond digits. But, actually dates in Javascript are only having milliseconds, not microseconds, so that is safe. I don't remember how we ended up with microseconds being stored in the DB, that triggered us to truncate precision. Configuring the precision at Knex level [is still not possible](https://github.com/knex/knex/issues/5308).

So, we had a way to soft delete before this PR, but it needed to be manual and was a real gotcha in our system, which may have caused some broken data. For example, if you hard delete an entity, there could be some leftovers due to weak references (`objectType` + `objectId` constructs we have in few tables), since they aren't foreign keys and would need to be deleted in code separately, which is often forgot. But, if you soft delete a record, it isn't actually deleted, so the links aren't really broken, the data is still there. If you however have unexpected behavior of the model, and hard delete a row in table that should be soft deleted, then you can have broken links in soft deletion tables too.

Having said all of that, I think you've done a good job in this PR to soft delete what should be soft deleted. But, I actually want to be able to hard delete records from soft deletion tables. It should be behind a flag (`forceHardDeletion`) of course and maybe add some warning documentation to warn about implications. I can currently see this option be used really rarely, but for example, I want to have an admin command to hard delete organizations, which are usually soft deleted (like via UI). This flag should only be offered for soft deletion tables. I would like if Typescript could infer that from value of `SoftDeletionEnabled` generic type, but I don't think it's possible. Instead, you'll likely need function overloads or similar thing to `AdditionalFindArgsForSequelizeTables`.

Also, don't forget to update the markdown documentation that I linked in my previous review.

src/db/util/sequelize-model.ts
```diff
@@ -72,6 +73,16 @@ export type AdditionalFindArgsForSequelizeTables = {
...
 };
 
+export type AdditionalDestroyArgsForSequelizeTables = {
+  /**
+   * If true, performs a hard deletion even when softDeletionEnabled is true.
+   * This permanently removes the record from the database.
+   *
+   * This is only available when softDeletionEnabled is true.
+   */
+  forceHardDeletion?: boolean;
```
Pl217: Let's tweak the docstring a bit to make it more scary when used
```suggestion
  /**
   * **THINK TWICE WHETHER YOU WANT THIS**
   * 
   * If `true`, performs a hard deletion even when `softDeletionEnabled` is `true`,
   * i.e. will hard delete records in tables which are otherwise soft deleted.
   *
   * This permanently removes the record from the database and only affects
   * tables with `softDeletionEnabled` set to `true`.
   */
  forceHardDeletion?: boolean;
```

Pl217 (changes requested):
You have a typo in last commit. Minor thing inline, otherwise this is good to go.

We can use `forceHardDeletion` even on tables which are not soft deleted:
```ts
await models.v4.plan.destroy({
  where: {
    restricted: true,
  },
  forceHardDeletion: true,
});
```
but that is also true for `includeDeleted`:
```ts
await models.v4.plan.find({
  where: {
    restricted: true,
  },
  includeDeleted: true,
});
```
I just tried these two snippets in hpc_service, having this branch checked out.

Pl217 (approved):
Just get the CI to pass

PR #224, opened 2025-02-26
Move plan deletion utils from hpc_service

Pl217:
In the past, I started an effort of doing all generic data operations in `hpc-api-core`, but I stopped doing that at some point. Now I remembered, so I move plan deletion helper code here from hpc_service

### DIFF
--- a/src/db/deletion/plan.ts
+++ b/src/db/deletion/plan.ts
@@ -1,0 +1,147 @@
+import type { Knex } from 'knex';
+import type { Database } from '..';
+import { deleteAuthTarget } from '../../auth';
+import { NotFoundError, UnprocessableEntityError } from '../../util/error';
+import type { ParticipantId } from '../models/participant';
+import type { PlanId } from '../models/plan';
+
+export const checkIfPlanIsDeletable = async (
+  database: Database,
+  planId: PlanId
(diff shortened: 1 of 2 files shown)


src/db/deletion/plan.ts
```diff
@@ -0,0 +1,145 @@
...
+   * `governingEntity`
+   *   - `governingEntityVersion`
+   *   - `globalClusterAssociation`
+   * `procedureEntityPrototype`
+   * `workflowStatusOption`
+   *   - `workflowStatusOptionStep`
+   * `planReportingPeriod`
+   *   - `measurement`
+   *     - `measurementVersion`
+   * `planVersion`
```
enxtur, line 131: Recently added **planVersionOrganization** table that you might want to include in this doc.

```suggestion
   * `planVersion`
   *   - `planVersionOrganization`
```

enxtur (changes requested):

src/db/deletion/plan.ts
Pl217: You are right and I have realized how terrible this is from maintenance standpoint. It will likely become outdated with any change.

enxtur (approved):

PR #233, opened 2025-03-06
Remove unnecessary usages of `any`

Pl217:
Since `Knex.Transaction` already defaults to `any` in its generic type definitions, us specifying them explicitly just floods our files with linter warnings about explicit usages of `any`.

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -498,7 +498,7 @@ export const deleteAuthTarget = async (
   database: Database,
   where: UserDataOfModel<Database['authTarget']>,
   actor: ParticipantId,
-  trx: Knex.Transaction<any, any>
+  trx: Knex.Transaction
 ): Promise<void> => {
   const authTargets = await database.authTarget.find({ where, trx });
   const authGrants = await database.authGrant.find({
--- a/src/db/deletion/project.ts
+++ b/src/db/deletion/project.ts
@@ -9,7 +9,7 @@ export const deleteProjectById = async (
   database: Database,
   projectToDelete: ProjectId,
   authGrantRevoker: ParticipantId,
-  trx: Knex.Transaction<any, any>
+  trx: Knex.Transaction
 ) => {
   const project = await database.project.findOne({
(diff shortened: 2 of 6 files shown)


Onitoxan (changes requested):
All these changes make sense and are good to go. Since I saw you needed to rebase I would propose also a few changes, that could belong to this PR:

https://github.com/UN-OCHA/hpc-api-core/blob/bf657e85a64c48da4e8f363c67bb39a4a4a14765/src/util/index.ts#L72-L73
to:
```ts
 export type PromiseType<P extends Promise<unknown>> = 
   P extends Promise<infer V> ? V : never; 
```
---
https://github.com/UN-OCHA/hpc-api-core/blob/bf657e85a64c48da4e8f363c67bb39a4a4a14765/src/util/index.ts#L200-L201
to:
```ts
export type MapValue< M extends Map<unknown,unknown>> =
  M extends Map<unknown, infer V> ? V : unknown;
 ```
---
https://github.com/UN-OCHA/hpc-api-core/blob/bf657e85a64c48da4e8f363c67bb39a4a4a14765/src/util/index.ts#L220-L221
to:
```ts
// eslint-disable-next-line @typescript-eslint/no-empty-object-type
export interface JSONArray extends Array<JSONValue> {}
```
---
https://github.com/UN-OCHA/hpc-api-core/blob/bf657e85a64c48da4e8f363c67bb39a4a4a14765/src/db/util/validation.ts#L36
to:
```ts
    .columnInfo()) as unknown as KnexColumns;
```

I thought we could remove some unnecessary `any` along thee way and also to fix an `eslint` disable comment that was updated.

Once you rebase you decide if you want to apply any of the proposed changes and as soon I see updates in this PR I approve 👍

Pl217:
Thanks @Onitoxan! That's the kind of review that I want to see! I have done what you suggested, but in the last commit, I am not changing the rule we suppress like you wanted, but removing that interface completely. The commit messages has more details.

Onitoxan (approved):

PR #234, opened 2025-03-07
Check formatting & lint files with `.cjs` & `.mjs` extensions

Pl217:
Besides what's in the title, I also slipped in two more changes:
- In PR #232, when we moved to ESM version of repo tools, we missed changing the names of config files
- The removal of unused eslint-disable directive could have been slipped in PR #233, since it already did the similar thing

### DIFF
--- a/src/db/util/sequelize-model.ts
+++ b/src/db/util/sequelize-model.ts
@@ -1,7 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-/*
- * Use of `any` in this module is generally deliberate to help with generics
- */
 import merge = require('lodash/merge');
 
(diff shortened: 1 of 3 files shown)


src/db/util/sequelize-model.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 /*
  * Use of `any` in this module is generally deliberate to help with generics
  */
```
Onitoxan: I would remove this comment since we have no explicit `any` in this file

Onitoxan (changes requested):
Just one inline comment to remove some unnecessary comments.

Onitoxan (approved):

PR #237, opened 2025-03-24
HPC-10007: Breakdown of requirements per org is missing for 2018 plans in planSummary and RMA API

enxtur:


src/db/models/endpointLog.ts
```diff
@@ -34,6 +34,7 @@ export const ENDPOINT_LOG_ENTITY_TYPE = t.keyof({
   procedureSection: null,
   project: null,
   'project comments': null,
+  'project segments': null,
```
Pl217, line 37: Don't add to this table. I am contemplating about removing `endpointLog` table

Pl217 (changes requested):
The reason data from 2018 is causing trends endpoint to return nothing is because data is broken. I have warned the clients regarding this when the API was initially developed and they assured me they only need recent plans and 2018 data is "too old".

Since now they want old data as well, the proper approach would be to fix the data itself and not change the code to account for broken data. We need to have strict data structure of budget segments, but there are plenty of problems we need to resolve, which is why I have opened an epic ticket [HPC-10160](https://humanitarian.atlassian.net/browse/HPC-10160).

To resolve the problem reported in [HPC-10007](https://humanitarian.atlassian.net/browse/HPC-10007), we should do the data fixing explained in [HPC-10161](https://humanitarian.atlassian.net/browse/HPC-10161). This PR should be closed.

enxtur:
Closing as incorrect assumption

PR #239, opened 2025-03-28
HPC-10144: Remove usages of `node-fetch`

Pl217:


src/auth/hid.ts
```diff
@@ -55,7 +51,7 @@ export const getHidInfo = async (
   } else {
     const accountUrl = new URL('/account.json', config.authBaseUrl);
     // Reference fetch.default to allow for mocking
```
enxtur: Looks like you missed this comment.

enxtur (approved):

src/auth/hid.ts
Pl217: Thanks for pointing this out

PR #240, opened 2025-03-31
Add array chunking utility method

Pl217:
This util method is moved here [from hpc_service](https://github.com/UN-OCHA/hpc_service/blob/0db71b2516e1f62bd11ff13076712b841855a084/app/commands/util/common.ts#L76-L82). Once we merge + release this code, we should replace usages in hpc_service and delete the definition from there.

Also defined a Solr & PostgreSQL constants that will likely be usually used with chunking method, but these weren't defined globally, just in [some temporary admin commands](https://github.com/UN-OCHA/hpc_service/blob/0db71b2516e1f62bd11ff13076712b841855a084/app/commands/commands/temp-clean-broken-plan-links.ts#L15) that I've been adding lately. Notice that I was using 2^15 in there, because I wanted to be on the safe side, but theoretical limit is 2^16 - 1. If you think we should play it safe, we can define 2^15 here too.

### DIFF
--- a/src/util/consts.ts
+++ b/src/util/consts.ts
@@ -1,0 +1,14 @@
+/**
+ * Maximum number of clauses allowed when parsing a boolean
+ * query string in Solr; 1024 is the default value
+ *
+ * [Source](https://solr.apache.org/guide/8_1/query-settings-in-solrconfig.html#maxbooleanclauses)
+ */
+export const SOLR_MAX_BOOLEAN_CLAUSES = 1024;
+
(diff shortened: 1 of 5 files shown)


enxtur (changes requested):
Needs testing :)

Pl217:
You are absolutely right! Thanks for making me write tests. I added test coverage for `range` utility method too

enxtur (approved):
Nicely done :)

PR #260, opened 2025-05-02
Return array instead of `Set` from `groupObjectsByProperty()`

Pl217:
There is a coincidence that we've upgraded to Node.js 22, so we can now use [`Map.groupBy()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy), which produces an array in `Map` values, rather than `Set` we were using. This newly available method isn't why I am making this change. The reason is that we've been using `Set` to no purpose. Using `Set` logically implies uniqueness of elements, but since this method is for grouping objects, they will be considered same only if:
> both are the same object (meaning both values reference the same object in memory)

I got this from definition of [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#description) because that is what `Set`s use for [comparing value equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#value_equality). `Set` actually uses SameValueZero algorithm, bit it is functionally the same for objects as Same-value equality, which is using `Object.is()`.

This can be illustrated with an example:
```js
new Set([{ duplicate: true }, { duplicate: true }]);
// Gives Set(2) {{…}, {…}}
```
The two objects are considered as different, even though they are same by value. As said, uniqueness of objects is only if they use same reference:
```js
const obj = { duplicate: true };
new Set([obj, obj]);
// Gives Set(1) {{…}}
```
Therefore, to break the illusion of uniqueness that is logically associated with `Set`s, the grouping is changed to arrays. This enables using `Map.groupBy()` and not the other way around. Method `groupObjectsByValue()` is removed, because it was essentially a polyfill for `Map.groupBy()`.

Using `Set`s wasn't only logically deceiving, but it also meant that we cannot easily iterate the result values, since `Set` interface was mostly basic. That has changed now and we have iterator helpers in ES2025, but they are lazy and we still prefer arrays, due to them being more logically correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unocha/hpc-api-core",
-  "version": "13.0.1",
+  "version": "14.0.0",
   "description": "Core libraries supporting HPC.Tools API Backend",
   "license": "Apache-2.0",
   "private": false,
--- a/src/lib/data/organizations.ts
+++ b/src/lib/data/organizations.ts
@@ -41,9 +41,9 @@ export const getOrganizationsInfo = async (
     await database.category.find({
       where: {
         id: {
-          [database.Op.IN]: [...categoriesByOrganizationId.values()]
-            .flatMap((crSet) => [...crSet])
-            .map((cr) => cr.categoryID),
+          [database.Op.IN]: categoriesByOrganizationId
+            .values()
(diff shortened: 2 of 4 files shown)


enxtur (approved):
Looks good to me

PR #263, opened 2025-05-09
Extend `JSON.stringify()` and `JSON.parse()`

Pl217:
Methods `JSON.stringify()` and `JSON.parse()` are extended to support serialization/deserialization of `Map` and `Set` built-ins.

Also, method `areSetsEqual()` is covered in unit tests now. This should have been done in same work when this method was added, but since it was used to test extended `JSON.stringify()` and `JSON.parse()`, it is also covered with tests here.

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -1,3 +1,4 @@
+import * as t from 'io-ts';
 import type { NonNegativeInteger } from './types';
 
 export const isDefined = <T>(v: T | null | undefined): v is T =>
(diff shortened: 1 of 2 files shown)


enxtur (approved):
Looks good to me.

In hindsight, I’m thinking that areSetsEqual might be somewhat inefficient—it computes the full symmetric difference just to check for equality.

Pl217:
> In hindsight, I’m thinking that areSetsEqual might be somewhat inefficient—it computes the full symmetric difference just to check for equality.

There is no built-in method to test equality in `Set` proposal, but there was [a discussion](https://github.com/tc39/proposal-set-methods/issues/69) why it wasn't added. My goal was to add symmetric difference method in the past and when new `Set` methods became available, the first thing I thought of when calculating difference between sets was symmetric difference. Combine that with my goal to have this as one liner and as short as possible, you've got what you got.

There are several other ways to check this:
1. `a.size === b.size && a.size === a.union(b).size`
2. `a.isSupersetOf(b) && a.isSubsetOf(b)`
3. `a.symmetricDifference(b).size === 0`
4. `a.size === b.size && a.isSubsetOf(b)` (this was [suggested](https://github.com/tc39/proposal-set-methods/issues/69#issuecomment-1303042740) in the discussion above)

As said, I went with 3rd option because it's shortest and also how I view calculating difference logically. I have done some benchmarking now and the ranking from fastest to slowest is 1-4-3-2, with 1 & 4 being really close in terms of performance and 2 & 3 are also really close between themselves, but more than twice slower than 1 & 4.

Between 1 & 4, I would prefer 4 because I see it as more readable, so if you want, I can open a PR to make this change.

enxtur:
@Pl217 Was the first option actually the fastest? I was initially thinking of something like this. But at this point, I’d prefer using the fourth option for me—it’s much more concise.
```ts
function areSetsEqual(a, b) {
  if (a.size !== b.size) return false;
  for (const item of a) {
    if (!b.has(item)) return false;
  }
  return true;
}
```

Pl217:
Yes, the first option was the fastest. The snippet you wrote was a valid option before we got new `Set` methods, but now we can write it more concise.

PR #264, opened 2025-05-09
Add io-ts codec to represent built-in `Map` type

Pl217:
Borrowed code from [here](https://github.com/gcanti/io-ts/issues/491#issuecomment-657436303) and modified slightly. Also made it throw an error if value is not a `Map`, because I think that is the correct behavior. Unit tests written with a help of LLM.

### DIFF
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -1,3 +1,6 @@
+import { isRight } from 'fp-ts/lib/Either';
+import * as t from 'io-ts';
+import { PathReporter } from 'io-ts/PathReporter';
 import type { FieldDefinition } from '../db/util/model-definition';
 
 /**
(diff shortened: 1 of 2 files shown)


src/util/types.ts
```diff
@@ -56,3 +59,50 @@ export const getTableColumns = <
...
+ * ```
+ */
+export const map = <K, V>(
+  domain: t.Type<K>,
+  codomain: t.Type<V>,
+  name = `Map<${domain.name}, ${codomain.name}>`
+): t.Type<Map<K, V>> => {
+  return new t.Type(
+    name,
+    (u): u is Map<K, V> => u instanceof Map,
```
enxtur: `instanceof` check is not enough. It should validate same as validator of `.decode`
```suggestion
    (u): u is Map<K, V> =>
      u instanceof Map &&
      [...u.entries()].every(
        ([key, value]) => domain.is(key) && codomain.is(value)
      ),
```

tests/utils/types.spec.ts
```diff
@@ -32,3 +34,71 @@ describe("Test 'getTableColumns' function", () => {
...
+    expect(isRight(result)).toBe(false);
+    if (!isRight(result)) {
+      expect(
+        result.left[0].context.some((c) => c.type.name === 'CustomMap')
+      ).toBe(true);
+      expect(result.left[0].message).toBe(
+        "Invalid type for key: '123'. Error: Invalid value 123 supplied to : CustomMap"
+      );
+    }
+  });
```
enxtur, line 103:
```suggestion
  });

  it('should validate the map', () => {
    expect(
      numberToStringMap.is(
        new Map([
          [1, 'a'],
          [2, 'b'],
        ])
      )
    ).toBe(true);
    expect(
      numberToStringMap.is(
        new Map([
          [1, 1],
          [2, 2],
        ])
      )
    ).toBe(false);
  });
```

enxtur (changes requested):

enxtur (approved):

PR #265, opened 2025-05-09
Add type-safe methods to convert `Map` to object and iterate it

Pl217:
I've added extensive docstrings with examples, which I usually don't do. So, I will not repeat myself here. The idea for adding these methods stemmed from desire to be able to convert `Map` to array and keep type safety of the keys, because `Object.fromEntries()` doesn't provide such type safety. Later, when such object (or any other) is iterated, the keys again need to be strongly typed, because `Object.entries()` is generic with `string` as key type (as both of them are actually).

**Edit**: After opening this PR, I have realized that we have `definedEntries()` method, which is very also a type-safe variant of `Object.entries()`, but it does a bit more (I mean the filtering, not unnecessary destructuring :) ). After learning this, I have moved my code to be in the same file, where the existence of both is more obvious.

There is also `mapObjectEntries()` in that file, but it does something else, even though it uses both `Object.fromEntries()` and `Object.entries()`, but it makes me wonder if `Object.fromEntries()` is meant to be used only to assemble back (processed results of) `Object.entries()` and if usage with other `Iterable` instances, like `Map` isn't the original idea, since their naming kinda indicates so.

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -12,6 +12,46 @@ export const definedEntries = <K extends string, V>(o: {
 }): Array<[K, V]> =>
   [...(Object.entries(o) as Array<[K, V]>)].filter(([_k, v]) => isDefined(v));
 
+/**
+ * Strict version of `Object.entries()` that has a more useful key type,
+ * but doesn't do filtering like `definedEntries()`.
+ *
+ * If we have object like this:
+ *
+ * ```ts
+ * const object = {
+ *   planLocation: [{ id: 1, planId: 100 }],
+ *   planYear: [{ id: 1, planId: 200 }],
+ * };
+ * ```
+ *
+ * looping over it using built-in `Object.entries()` with
+ *
+ * ```ts
+ * for (const [key, value] of Object.entries(object)) { ... }
+ * ```
+ *
+ * will produce the type of `key` as generic `string` and type of `value` is
+ *
+ * ```ts
+ * {
+ *   id: number;
+ *   planId: number;
+ * }[] | {
+ *   id: number;
+ *   planId: number;
+ * }[]
+ * ```
+ *
+ * If we were to use loop over `objectEntries(object)`, with this
+ * method instead of built-in `Object.entries()`, we would get
+ * the type of `key` as `"planLocation" | "planYear"` and type of `value`
+ * stays the same, as it was already properly typed.
+ */
+export const objectEntries = <K extends PropertyKey, V>(obj: {
+  [P in K]: V;
+}): Array<[K, V]> => Object.entries(obj) as Array<[K, V]>;
+
 export const delay = (ms: number): Promise<unknown> =>
   new Promise((resolve) => setTimeout(resolve, ms));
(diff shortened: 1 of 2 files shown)


Onitoxan (approved):
Is good to go! Honestly is a great idea to add this method. I was trying to see if we could go even one step further but I think we are limited by TS. `Object.entries()` will actually make a predictable result since it follows this ordering rules:

1. Integer keys are listed in ascending numeric order.
2. `string` keys are listed in the order they were added.

I tried to modify the code to get more predictable results but didn't find a way. But I'll write down what I think in case you see a way to do it, even though I think we can't

```ts
const foo = {a:1, b:2 ,4:6}
const res = objectEntries(foo)
foo[0][1] // The real type of this should be `6` instead of `1 | 2 | 6`
```

Anyways, is good to go and merge, just let me know if you come up with a way to do what I mention.

Pl217:
@Onitoxan In that code, you actually have a type error. Maybe you wanted to use index accessors on `res`, not on `foo`? If third line is `res[0][1]`, then type is just `number`, because type of `foo` is:
```ts
{
  a: number;
  b: number;
  4: number;
}
```

Onitoxan:
@Pl217 Yeah, didn't spend much time writing the example, I meant accessing to `res` and I also missed to write on `foo` definition: 

```ts
const foo = {a:1, b:2 , 4:6} as const
```
what would make
```ts
{
  a: 1;
  b: 2;
  4: 6;
}
```
Instead of `number`

Pl217:
Yes, I tried with `as const` too, but real type of `res[0][1]` should still not be `6` as you wrote. What you want shouldn't be possible, because Javascript objects don't have guaranteed order and Typescript doesn't either. Arrays guarantee orders, so if we have this:
```ts
const array = [
  ['a', 1],
  ['b', 2],
  ['c', 3],
] as const;
```
its type would be `readonly [readonly ["a", 1], readonly ["b", 2], readonly ["c", 3]]`, but even if I rewrite `objectEntries()` to work with readonly array of readonly tuples, it will still be an array (not a tuple) created from an object, whose keys order we cannot know and it isn't guaranteed. So, with this method:
```ts
const objectEntries = <T extends Record<string, unknown>>(
  obj: T
): ReadonlyArray<{ [K in keyof T]: readonly [K, T[K]] }[keyof T]> => {
  return Object.entries(obj) as any;
};
```
if I write this code
```ts
const obj = {
  a: 1,
  b: 2,
  c: 3,
} as const;

const entries = objectEntries(obj);
```
then `entries` will be typed as `readonly (readonly ["c", 3] | readonly ["a", 1] | readonly ["b", 2])[]` or any variation, it could be in any order. Compare this to above readonly tuple type to see the difference.

I don't think your conclusion that "Integer keys are listed in ascending numeric order" and "string keys are listed in the order they were added" is correct, as I've seen all variations while playing around with it. So, in my example, `entries[0][1]` could be inferred as `2 | 3 | 1` or `1 | 2 | 3`, or presumably any other variation, since order is not guaranteed.

PR #269, opened 2025-05-13
HPC-9642: Add new model 'ConfigurationField'

manelcecs:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@unocha/hpc-api-core",
-  "version": "14.1.0",
+  "version": "15.0.0",
```
Onitoxan: This is always a struggle for me, honestly I'm not 100% sure, but I think this should be `v14.2.0` because we are not introducing a breaking change, so since it's a new "functionality" that is retro-compatible I think it should be a minor version increase. I don't want to be the semver police though, so let me know what do you think.

Onitoxan (changes requested):

package.json
manelcecs: Makes sense!

Onitoxan (approved):

src/db/models/configurationField.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type ConfigurationFieldId = Brand<
+  string,
+  { readonly s: unique symbol },
+  'configurationField.id'
+>;
+
+export const CONFIGURATION_FIELD_ID = brandedType<string, ConfigurationFieldId>(
+  t.string
+);
```
Pl217, line 15: I think this should be named not to mention ID, since IDs are numeric sequences and you're putting a string primary key on column `name` here.  This should have been:
```ts
export type ConfigurationFieldName = Brand<
  string,
  { readonly s: unique symbol },
  'configurationField.name'
>;

export const CONFIGURATION_FIELD_NAME = brandedType<
  string,
  ConfigurationFieldName
>(t.string);
```
And this is only if you really want to brand the string. The branding was added primarily for numerical IDs, to prevent all IDs being just plain `number`, which would create problems because it would allow searching by ID in one table using IDs from another table, which is obviously wrong. Branding was added to prevent making these bugs, which is great as we'd make many bugs without branded IDs.

But for strings, your intent would be to differentiate just plain string literal that can mean "anything" from a string that is actually a primary key of this table. Do you think we need that? With that, you'd need to brand strings when searching this table, like this:
```ts
await models.v4.configurationField.find({
  where: {
    name: createBrandedValue('someValueRepresentingPrimaryKeyID'),
  },
});
```

This is the only table where we have string primary key ID, so branding the string would not be so useful in terms of confusing the primary keys between tables. Table `iatiPublisher` also has string primary key ID, but it isn't used anywhere, should be removed in [HPC-10137](https://humanitarian.atlassian.net/browse/HPC-10137) and is poorly defined actually. I would fix that, if the table wasn't dead code.

Pl217 (commented):
Why was the wrong ticket in the head branch name? 😄

package.json
Pl217: @Onitoxan is right. Adding a new model would mean a minor version bump, not major

PR #270, opened 2025-05-16
Fix `type` column potential value in `expiredData` model

Pl217:
The [schema change](https://github.com/UN-OCHA/hpc_service/blob/f7093b1146b573807fbff3dc49ee7ec3c23d11a7/app/models/migrations/z.release3.12.07.v001.migration.ts#L9-L11) which introduced this table clearly indicates that `type` column can only have value `'flow'` and not `'solr'`.

Other column named `objectType` can be `'flow'` or `'project'`, so it was potentially a mistake in the enum in that schema change and `'solr'` was actually supposed to be a value, but since table isn't used for anything, I just changed model to make it possible to use this table when writing unit tests. If you think that changing the enum in the DB instead of changing the model would be better, I can do that too.

### DIFF
--- a/src/db/models/expiredData.ts
+++ b/src/db/models/expiredData.ts
@@ -3,7 +3,7 @@ import { DATE } from '../util/datatypes';
 import { defineRawModel } from '../util/raw-model';
 
 const EXPIRED_DATA_TYPE = {
-  solr: null,
+  flow: null,
 };
 
 const EXPIRED_OBJECT_TYPE = {


enxtur (commented):
From the related tickets, I’ve gained a partial understanding of the purpose of the expiredData table. However, I couldn’t find where the data is being inserted into it. Is it used by some external service? Do we still need this table?

Pl217:
> From the related tickets, I’ve gained a partial understanding of the purpose of the expiredData table. However, I couldn’t find where the data is being inserted into it. Is it used by some external service? Do we still need this table?

This table was part of a bigger work where we wanted to have a table to mark which data has expired, because data deletions were not reflected in Solr. The work was split into smaller tickets and one of the first tasks was to create a table, which was done. But, subsequent tickets to make use of it and clean Solr regularly wasn't done as developer who was working on this left the team and no one was assigned the work afterwards.

I think it's still important, so we haven't removed the table, with hope to someday resolve the problem of Solr cleanup. You made me look at the original ticket [HPC-8079](https://humanitarian.atlassian.net/browse/HPC-8079) and in there I see that `"solr"` was the intended value for `type`, so I'll change the enum in the database, because that is where mistake is, not in the codec here.

[HPC-8079]: https://humanitarian.atlassian.net/browse/HPC-8079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #280, opened 2025-06-19
Improve organization deletion error message

Pl217:
The original intent was to truncate number of IDs shown in error message, because some organizations have tens of thousands of associated flows/projects, so the message gets ridiculously long. The slowness was observed in those cases, so I tried to optimize, by making fewer queries, since we were fetching flows using `findOne()`, which is too many queries when organization has tens of thousands of flows.

Antonio also asked to display project IDs instead of project version IDs, so I've done that as well. In that case, I was being optimal, by only fetching 10 that we'll display, but in case of fetching flows, I could not have done that, since they can be soft deleted, so querying `flow` table filters those out.

If you want to see the organizations with lots of flow connections, use this query:
```sql
SELECT fo."objectID", COUNT(*)
FROM "flowObject" fo
GROUP BY fo."objectID", fo."objectType"
HAVING (fo."objectType" = 'organization' OR fo."objectType" = 'anonymizedOrganization')
ORDER BY COUNT(*) DESC;
```

If you want to check how many unique flow versions there are (because above doesn't filter out deleted flows and situations where organization is on both source and destination side of same flow), then use this query (e.g. for organization ID 2915):
```sql
SELECT DISTINCT fo."flowID", fo."versionID"
FROM "flowObject" fo
JOIN flow f ON
  fo."flowID" = f.id AND fo."versionID" = f."versionID"
WHERE
  fo."objectID" = 2915 -- CHANGE THIS
  AND (
    fo."objectType" = 'organization' OR 
    fo."objectType" = 'anonymizedOrganization'
  )
  AND f."deletedAt" IS NULL
ORDER BY
  fo."flowID" DESC,
  fo."versionID" ASC;
```

### DIFF
--- a/src/db/deletion/organization.ts
+++ b/src/db/deletion/organization.ts
@@ -83,10 +83,13 @@ export const deleteOrganizationById = async (
           })
       )
     )
-  ).flat();
+  )
+    .flat()
+    .toSorted((firstFlow, secondFlow) => secondFlow.id - firstFlow.id);
 
   const projectVersions = await database.projectVersionOrganization.find({
     where: { organizationId: organization.id },
+    orderBy: { column: 'projectVersionId', order: 'desc' },
     trx,
   });
 
--- a/src/db/deletion/organization.ts
+++ b/src/db/deletion/organization.ts
@@ -55,17 +55,27 @@ export const deleteOrganizationById = async (
   });
 
   if (flows.length > 0 || projectVersions.length > 0) {
-    const flowIds = flows.map((flow) => flow.id).join(', ');
-    const projectVersionIds = projectVersions
-      .map((project) => project.projectVersionId)
-      .join(', ');
+    const formatIds = (ids: number[], limit: number = 10) => {
+      if (ids.length <= limit) {
+        return ids.join(', ');
+      }
+
+      const displayedIds = ids.slice(0, limit).join(', ');
+      const remainingCount = ids.length - limit;
+      return `${displayedIds} ... (+ ${remainingCount.toLocaleString()} more)`;
+    };
+
+    const flowIds = flows.map((flow) => flow.id);
+    const projectVersionIds = projectVersions.map(
+      (project) => project.projectVersionId
+    );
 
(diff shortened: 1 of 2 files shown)


src/db/deletion/organization.ts
```diff
@@ -26,46 +65,74 @@ export const deleteOrganizationById = async (
...
+
+      const displayedIds = ids.slice(0, limit).join(', ');
+      const remainingCount = ids.length - limit;
+      return `${displayedIds} ... (+ ${remainingCount.toLocaleString()} more)`;
+    };
+
+    const flows = flowVersions.map((flow) => `${flow.id} v${flow.versionID}`);
+    const projectIds = [
+      ...projectVersions.map((pv) => pv.projectId.toString()),
+      ...restProjectVersions.map((pvo) => pvo.projectVersionId.toString()),
```
enxtur, line 128: @Pl217 Is it intentional? I thought you wanted to show project IDs.

Pl217: Yes, this is intentional, as explained in the [comment above](https://github.com/UN-OCHA/hpc-api-core/blob/ef43d7390cd39b682f9890737779b8f34d9aca35/src/db/deletion/organization.ts#L96-L98). Since we're going to display only 10 project IDs, I didn't want to fetch all of them and I use project versions just to get the total count, since error message will list 10 project IDs and then truncate the rest, displaying the number of remaining projects (for which I use versions). The example error message is:
> Cannot delete organization. Associated flow versions: [347552 v1, 347551 v2, 347551 v1, 347550 v2, 347550 v1, 347549 v2, 347549 v1, 347548 v2, 347548 v1, 347547 v2 ... (+ 22,523 more)]. Associated projects: [228271, 227173, 227168, 225623, 224722, 224684, 224575, 224338, 224117, 223636 ... **(+ 1,950 more)**].

Since this wasn't clear, we should improve the situation. I'd add one more comment here saying `// As said above, for the remaining count, we use project version IDs`, but if you have a better idea, let me know.

enxtur (approved):

PR #293, opened 2025-07-07
HPC-10197: Enhance legacy versioned model deletion

enxtur:


src/db/util/sequelize-model.ts
```diff
@@ -218,6 +218,24 @@ export const defineSequelizeModel =
...
+            ...destroyArgs,
+            values: {
+              ...values,
+              currentVersion: false,
+              latestVersion: false,
+              latestTaggedVersion: false,
+            },
+          });
+          return res.length;
+        }
```
Pl217: This should have been done in "legacy versioned model" (file `legacy-versioned-model.ts`) and not here.

Pl217 (changes requested):
So, our soft deletion didn't properly set these properties to `false`. Do we need to write a script to fix the affected rows?

src/db/util/legacy-versioned-model.ts
```diff
@@ -42,23 +47,59 @@ export type FieldsWithVersioned<
...
+        latestVersion: false,
+        latestTaggedVersion: false,
+      };
+      if (
+        opts.softDeletionEnabled &&
+        !shouldForceHardDeletion &&
+        isSoftDeletionValues<FieldsWithVersioned<F, SoftDeletionEnabled>>(
+          values
+        )
+      ) {
```
Pl217: Did you test this? I think it will fail to execute because `isSoftDeletionValues()` checks that `values` is object with single key `deletedAt`, because you have this logic `Object.keys(values).length === 1 && 'deletedAt' in values`

Pl217 (changes requested):
I thought of approach where `legacy-versioned-model` updates the fields it defines (`currentVersion`, `latestVersion` and `latestTaggedVersion`) and doesn't deal with `sequelize-model`, which should keep setting its own `deletedAt`. Even though legacy model extends sequelize model, it doesn't have to know the internals of sequelize model and can only deal with its own properties, delegating the rest to sequelize model and below.

This is the code that I came up with. The downside is that I'm updating `current`/`latest`/`latestTagged` properties even when the row will be hard deleted. Also, I got away from defining a type guard by using type assertion with `as`.
```ts
import merge = require('lodash/merge');
import * as t from 'io-ts';

import {
  defineIDModel,
  type FieldsWithId,
  type ModelWithIdInitializer,
} from './id-model';
import type { FieldDefinition, UserDataOf } from './model-definition';
import type { DestroyFn } from './raw-model';

const VERSIONED_FIELDS = {
  nonNullWithDefault: {
    currentVersion: {
      kind: 'checked',
      type: t.boolean,
    },
    latestVersion: {
      kind: 'checked',
      type: t.boolean,
    },
    latestTaggedVersion: {
      kind: 'checked',
      type: t.boolean,
    },
    versionTags: {
      kind: 'checked',
      type: t.array(t.string),
    },
  },
} as const;

type ExtendedFields<F extends FieldDefinition> = F & typeof VERSIONED_FIELDS;

export type FieldsWithVersioned<
  F extends FieldDefinition,
  SoftDeletionEnabled extends boolean,
> = FieldsWithId<F, SoftDeletionEnabled> & ExtendedFields<F>;

/**
 * A model that has been defined using `versionLib.versionModel`
 *
 * This definition function extends the given table definition with columns that
 * are present on all models defined by the function `versionLib.versionModel`.
 */
export const defineLegacyVersionedModel =
  <
    F extends FieldDefinition,
    IDField extends string &
      (keyof F['generated'] | keyof F['generatedCompositeKey']),
    SoftDeletionEnabled extends boolean,
  >(opts: {
    tableName: string;
    fields: F;
    idField: IDField;
    softDeletionEnabled: SoftDeletionEnabled;
  }): ModelWithIdInitializer<
    FieldsWithVersioned<F, SoftDeletionEnabled>,
    IDField
  > =>
  (masterConn, replicaConn) => {
    const fields: ExtendedFields<F> = merge({}, opts.fields, VERSIONED_FIELDS);

    const model = defineIDModel({
      ...opts,
      fields,
    })(masterConn, replicaConn);

    const destroy: DestroyFn<
      FieldsWithVersioned<F, SoftDeletionEnabled>
    > = async (args) => {
      const values = {
        currentVersion: false,
        latestVersion: false,
        latestTaggedVersion: false,
      } as Partial<UserDataOf<FieldsWithVersioned<F, SoftDeletionEnabled>>>;

      await model.update({
        ...args,
        values,
      });

      return await model.destroy(args);
    };

    return { ...model, destroy };
  };

```

Pl217 (approved):
You didn't answer my question whether existing data needs fixing, since our soft deletion didn't properly set version properties to `false`. Do we need to write a script to fix the affected rows?

enxtur:
@Pl217 I’m not keen on double updates, but I like more the idea of handling its own properties. For affected data, I didn't dig deep. We sure can make some script since it's easy to implement.

PR #316, opened 2025-10-10
Check existing email for first HID login too

Pl217:
The existing code was written based on assumption that users switching to Entra ID already used HID before, so if the same email was used for HID that is now used for Entra ID, the same record in `participant` table would be kept.

But, that assumption is not necessarily true. While most users would have used HID before switching to Entra ID, it doesn't have to mean that they used same email address for both.

The fix wants to improve the situation if users first log into our system with Entra ID, then create HID with that same email address, regardless if they already have HID accounts with other email addresses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unocha/hpc-api-core",
-  "version": "15.0.0",
+  "version": "15.0.1",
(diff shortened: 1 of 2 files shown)


enxtur (approved):
LGTM!

